PR #27, opened 2014-06-06
gp.sh uses http instead of git protocol

cletusw:
Don't know if this affects anyone else, but the git protocol is sometimes blocked by firewalls for me.

### DIFF
--- a/bin/gp.sh
+++ b/bin/gp.sh
@@ -20,7 +20,7 @@ branch=${3:-"master"} # default to master when branch isn't specified
 
 # make folder (same as input, no checking!)
 mkdir $repo
-git clone git@github.com:$org/$repo.git --single-branch
+git clone https://github.com/$org/$repo.git --single-branch
 
 # switch to gh-pages branch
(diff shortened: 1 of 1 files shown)


ssorallen:
I'm in favor of HTTPS instead of SSH as well because GitHub suggests it. I opened a similar issue for the pull-all script: https://github.com/Polymer/tools/issues/44

bin/gp.sh
```diff
@@ -11,7 +11,7 @@ branch=${3:-"master"} # default to master when branch isn't specified
 
 # make folder (same as input, no checking!)
 mkdir $repo
-git clone git@github.com:$org/$repo.git --single-branch
+git clone http://github.com/$org/$repo.git --single-branch
```
ssorallen: This should probably be "https". GitHub redirects to https anyway, so that will save the script a redirect.

cletusw: Done :)

rictic:

PR #46, opened 2015-02-11
Include dev dependencies when seeding gh-pages branch

MaKleSoft:
Some dependencies like `core-component-page` are only used in demos or for displaying documentation. These components should not be included as regular dependencies in the `bower.json` because that will cause them to be unnecessarily installed when the parent component is used in another project. An alternative strategy is to include these dependencies as 'devDependencies' instead so they are only installed when in a dev environment. This pull request changes the way components are installed when seeding the gh-pages branch so that dev dependencies are included along with regular dependencies.


rictic:

PR #54, opened 2015-10-06
Add argument to specify remote URL

gazal-k:

PR #59, opened 2016-04-20
Use explicit url for bower install of host element. Fixes #58

### LINKED ISSUE #58: gp.sh not working for private repos
Running `gp.sh` for a non-public repository fails with the following error:

``` sh
bower my-element#master       not-cached git://github.com/my-org/my-element.git#master
bower my-element#master          resolve git://github.com/my-org/my-element.git#master
bower my-element#master          ECMDERR Failed to execute "git ls-remote --tags --heads git://github.com/my-org/my-element.git", exit code of #128 fatal: remote error:    Repository not found.
```

The problem here seems to be the repository url `git://github.com/my-org/my-element.git#master` which should probably be something like `git@github.com:my-org/my-element.git`.

MaKleSoft:


TimvdLippe:
It seems that #54 also fixes this issue.

rictic:

PR #60, opened 2016-04-30
Clean up directory after gp.sh has finished

TimvdLippe:
When running the script, it leaves behind the directory it created the `gh-pages` in. This change moves this to a `temp/` folder, as running the script also had problems when the directory already existed. After the script ran, remove the `temp/$repo` folder again.


TimvdLippe:
Seems like this repository is abandoned, so just closing this PR.

PR #65, opened 2017-03-31
Include notes for travis encryption issues

krumware:
This is quite silly, but if the local repository configuration does not have a case-sensitive match with Travis, encryption will fail.
ex: repo as `myOrg/my-component` in Travis, but cloned locally as `myorg/my-component` will cause error during encrypt commands: 
```
repository not known to https://api.travis-ci.org/: myorg/my-component
```
Attempting to help others with this "gotcha"


rictic:

PR #78, opened 2018-04-09
jsTransform now inlines Babel helpers by default.

aomarks:
Previously they never did. Added the `omitInlineHelpers` option to optionally disable inlining.

Addresses #987

### DIFF
--- a/packages/build/src/js-transform.ts
+++ b/packages/build/src/js-transform.ts
@@ -36,8 +36,6 @@ const babelPresetEs2015NoModules =
 // are small, and we have to list syntax plugins separately anyway (see below),
 // just enumerate the transform plugins here instead of merging the 3 presets.
 const babelTransformPlugins = [
-  // Don't emit helpers inline.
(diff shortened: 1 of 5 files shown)


packages/build/src/js-transform.ts
```diff
@@ -75,11 +76,15 @@ const babelPresetMinify =
...
-  // Note that some JavaScript features may require the Babel helper polyfills,
-  // which this function will not insert and must be loaded separately.
   compileToEs5?: boolean;
 
+  // If true, do not include Babel helper functions in the output. Otherwise,
+  // any Babel helper functions that were required by this transform (e.g. ES5
+  // compilation or AMD module transformation) will be automatically included
+  // inline with this output. If you set this option, you must provide those
+  // required Babel helpers by some other means.
+  omitInlineHelpers?: boolean;
```
justinfagnani: `externalHelpers`?

aomarks: Done.

PR #85, opened 2018-04-10
Pull recent cli commits

aomarks:
These were committed after the initial CLI monorepo import.


aomarks:
Previous merge was wrong. Was importing the entire polymer-cli into a new packages/polymer-cli (instead of polymer/cli). This one is fixed. PTAL.

rictic:
Ah, that explains the diff!

PR #92, opened 2018-04-10
Some cleanup that gets us incrementally closer to hoisting

justinfagnani:


packages/analyzer/package.json
```diff
@@ -78,7 +78,7 @@
     "@types/doctrine": "^0.0.1",
     "@types/is-windows": "^0.2.0",
     "@types/minimatch": "^3.0.1",
-    "@types/node": "^6.0.0",
+    "@types/node": "^9.6.4",
```
rictic: Do we want to use the newest supported major node version typings, or the oldest? My intuition is oldest.

justinfagnani: Some of the changes are just more accurate typings. Also, the problem goes both ways: older typings might still include a deprecated or removed API. We have to use common APIs no matter what...

rictic: True, but in node things are deprecated much slower than they're added. Ultimately though, I won't block on this.

PR #97, opened 2018-04-10
Fix build no package lock

aomarks:


rictic:
Why nuke all package lock files?

justinfagnani:
It's because of a bad interaction with lerna and package locks on windows. I added removing package locks to the AppVeyor config, so closing this one.

PR #101, opened 2018-04-11
Bump cli caret dep on polyserve in prep for 1.7.0-pre.7

aomarks:
This might not technically be necessary, but it is what lerna would have
done if we had published the two packages together. And it ensures users
get the recent fix to polyserve despite their package lock or whatever,
which is the main reason for this release.


rictic:
A little paranoid, but no objections

Issue #113, opened 2018-04-11
[cli] test command does not forward `moduleResolution` option in `polymer.json` to WCT


bicknellr:
The `--module-resolution` is not passed to WCT by `polymer test`: 
https://github.com/Polymer/tools/blob/ce2d856c48212caf7e41c96bbf9f658420c488c9/packages/cli/src/commands/test.ts#L142-L157
Users have to add it manually to test if they want WCT's server to transform name specifiers.

---

**Correction:** `polymer test` passes all of the command line flags but it does not add the `--module-resolution` flag when `moduleResolution` is set in `polymer.json`.

bicknellr:
Correction: `polymer test` passes all of the command line flags but it does not add the `--module-resolution` flag when `moduleResolution` is set in `polymer.json`.

PR #115, opened 2018-04-11
Adds a Polymer 3 element template to `init`.

bicknellr:


bicknellr:
This still doesn't install.

bicknellr:
I added an integration test, there were a handful of problems though:
- `polymer install` only tries toinstall bower packages so I had to manually run `npm install`. (#118)
- `polymer lint` doesn't have a Polymer 3 option.
- `polymer test` doesn't pass the `--module-resolution=node` flag to WCT. (#113)

packages/cli/src/init/element/element.ts
```diff
@@ -119,4 +119,51 @@ export function createElementGenerator(templateName: string):
           'Check out your new project README for information about what to do next.\n');
     }
   };
+
+  class Polymer3ElementGenerator extends ElementGenerator {
+    // TODO(bicknellr): Why doesn't this inherit properly? Because `async` is
+    // compiled out?
+    async prompting() {
+      return super.prompting();
+    }
```
rictic: Can you double check that this doesn't work when this isn't here? I looked at the JS that's produced, and we're producing ES6 here. I can't see any reason why this code would do anything at all.

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -34,6 +37,24 @@ suite('integration tests', function() {
...
+    skipOnWindows('test the Polymer 3.x element template', async () => {
+      const dir =
+          await runGenerator(createElementGenerator('polymer-3.x'))
+              .withPrompts({name: 'my-element'})  // Mock the prompt answers
+              .toPromise();
+      // TODO(bicknellr): Use `polymer install` once it supports installing npm
+      // packages.
+      await exec('npm install', {cwd: dir});
+
+      // TODO(bicknellr): Add this back in when `polymer lint` has a Polymer 3
```
rictic: Let's use test urls instead of usernames for TODOs. https://github.com/Polymer/tools/issues/130 for this one

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -34,6 +37,24 @@ suite('integration tests', function() {
...
+      // option.
+      // await runCommand(binPath, ['lint'], {cwd: dir});
+
+      // TODO(bicknellr): Remove the `--module-resolution=node` argument once
+      // `polymer test` passes them in correctly
```
rictic: https://github.com/Polymer/tools/issues/113

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -34,6 +37,24 @@ suite('integration tests', function() {
 
   suite('init templates', () => {
 
+    skipOnWindows('test the Polymer 3.x element template', async () => {
+      const dir =
+          await runGenerator(createElementGenerator('polymer-3.x'))
+              .withPrompts({name: 'my-element'})  // Mock the prompt answers
+              .toPromise();
+      // TODO(bicknellr): Use `polymer install` once it supports installing npm
```
rictic: #118

packages/cli/src/init/element/element.ts
bicknellr: It seems like async-await is being compiled out. I'm betting it has something to do with that.

bicknellr: Also, yes, the parent class' `prompting` function isn't called if this isn't here. I can't figure out where yeoman-generator calls it though and I'm not sure how it gets called due to the async-await thing.

bicknellr: Hmm, telling TypeScript to target ES2017 still has this problem.

bicknellr: Here's the problem: https://github.com/yeoman/generator/blob/v1.0.1/lib/index.js#L382
Yeoman only checks the object's prototype's own properties, so I think I have to have something like this in here.

bicknellr: I've added pass-through functions for all of the other methods as well.

rictic: Oh hey, you found it too. I just filed https://github.com/yeoman/generator/issues/1065

packages/cli/src/init/element/element.ts
```diff
@@ -121,8 +121,14 @@ export function createElementGenerator(templateName: string):
   };
 
   class Polymer3ElementGenerator extends ElementGenerator {
-    // TODO(bicknellr): Why doesn't this inherit properly? Because `async` is
-    // compiled out?
+    // This is not a no-op: Yeoman only checks the object's prototype's own
```
rictic: Let's add a mention of yeoman/generator#1065 here with these `not a no-op` comments.

bicknellr:
One of the tests timed out; I'm going to restart that Travis build.

PR #119, opened 2018-04-12
Fix wrong AMD transform dependency/callback parameter ordering.

### LINKED ISSUE #112: AMD transform does not update module callback arguments correctly
Our AMD transform injects a dependency on the previous module into the AMD module dependency list. However, it does not update the positional argument of the module callback. Example:

```js
define('polymer-build-generated-module-1', ['polymer-build-generated-module-0', "./node_modules/@polymer/polymer/polymer-element.js"], function (_polymerElement)
```

aomarks:
Fix AMD transform bug where if an HTML document had multiple type=module
scripts, and any of them (apart from the first) had any kind of import,
then that import was not accessible (because it was mapped to the wrong
module callback function argument).

Fixes #112

### DIFF
--- a/packages/build/CHANGELOG.md
+++ b/packages/build/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Fix AMD transform bug where if an HTML document had multiple type=module scripts, and any of them (apart from the first) had any kind of import, then that import was not accessible (because it was mapped to the wrong module callback function argument).
 <!-- Add new, unreleased changes here. -->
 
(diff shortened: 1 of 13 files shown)


packages/build/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Fix AMD transform bug where if an HTML document had multiple type=module scripts, and any of them (apart from the first) had any kind of import, then that import was not accessible (because it was mapped to the wrong module callback function argument).
```
rictic: What about a policy where we update all affected packages' changelogs when there's important bugfixes or new features? So in this case we'd copy this change notice to CLI and Polyserve too.

That'll also help `lerna publish` infer when it should stamp a new release of a package.

aomarks: Done.

PR #124, opened 2018-04-12
Add browser-capabilities to vscode config

aomarks:


tools.code-workspace
```diff
@@ -3,6 +3,9 @@
 		{
 			"path": "packages/analyzer"
 		},
+	        {
```
justinfagnani: Can you sort the whole list while you're here?

aomarks: Done. I don't actually even know what this file does.

rictic:
It's useful for expressing the project structure to vscode

PR #129, opened 2018-04-13
Let the ProjectConfig handle creating a configured analyzer and warning filter in the CLI

rictic:


packages/cli/src/lint/lint.ts
```diff
@@ -48,21 +48,28 @@ export async function lint(options: Options, config: ProjectConfig) {
...
 
   if (options.watch) {
     return watchLoop(
-        analyzer, urlLoader, urlResolver, linter, options, config, filter);
+        analyzer,
+        urlLoader,
+        urlResolver,
+        linter,
+        options,
+        config,
```
valdrinkoshi: Noticed that we pass around `config` but we only use `config.root`, would it make sense to refactor all these signatures to get the root instead?

rictic: Next time I touch this file I'm gonna refactor this into a class rather than pass these same 7-8 options around everywhere

PR #136, opened 2018-04-13
Fix travis definition of integration tests.

rictic:
Previous to this, our integration tests were only _actually_ running on appveyor, not on travis.


usergenic:
Okay it looks like the previous integration test failures were a hiccup of sorts.  Rerunning made them pass.  Approving and squashing and merging.

PR #144, opened 2018-04-14
Add common tsconfig-base.json and use in all packages

aomarks:
The idea is to have a strict base config in the top-level of tools/. Each package extends the base, sets relative source paths (which are always interpreted relative to the file that references them), and any required additional compiler flags.

The "extends" syntax requires gulp-typescript >= 3, so I did that upgrade in Analyzer.


packages/analyzer/gulpfile.js
```diff
@@ -83,7 +83,7 @@ task('compile', function () {
     gulp.src('src/**/*.ts');  //.pipe(newer({dest: 'lib', ext: '.js'}));
   const tsResult =
     srcs.pipe(sourcemaps.init())
-      .pipe(typescript(tsProject, [], typescript.reporter.fullReporter()));
+      .pipe(tsProject(typescript.reporter.fullReporter()));
```
rictic: Why this change?

Not that I'm complaining, but touching gulpfiles is usually so obnoxious that I assume someone has to have had a really good reason to have done so, so I'm curious :)

aomarks: There were some API changes going from gulp-typescript 2->4, which I needed because gulp-typescript 2 didn't support "extends".

PR #145, opened 2018-04-14
Adds support for npm to `polymer install`.

### LINKED ISSUE #118: [cli] Should `polymer install --npm` run `npm install`?
When running `polymer install` (with or without `--npm`) in a 3.0-style package with a `package.json` instead of a `bower.json`, this error is produced:
```
info: [cli.install]    Installing default Bower components...
error: [cli.main]   cli runtime exception: Error: No bower.json present
error: [cli.main]   Error: No bower.json present
    ... stack trace ...
```
Maybe we should make the `--npm` flag cause `polymer install` to run `npm install` if a `package.json` is found?

bicknellr:
If the `--npm` flag is passed or `npm: true` is set in `polymer.json`, `polymer install` will run `npm install` instead of installing dependencies from Bower. (fixes #118)


packages/cli/src/commands/install.ts
```diff
@@ -28,27 +28,34 @@ export class InstallCommand implements Command {
...
   ];
 
-  async run(options: CommandOptions, _config: ProjectConfig): Promise<void> {
+  async run(options: CommandOptions, config: ProjectConfig): Promise<void> {
     // Defer dependency loading until this specific command is run
     const install =
         require('../install/install').install as typeof installTypeOnly;
+
+    // Use `npm` from the config, if available and not passed as a CLI arg.
+    if (options.npm === undefined && config.npm !== undefined) {
```
rictic: We should really do this in a unified place

packages/cli/src/install/install.ts
```diff
@@ -14,14 +14,17 @@
...
 import * as path from 'path';
 import * as logging from 'plylog';
+import * as util from 'util';
 
 import defaultBowerConfig = require('bower/lib/config');
 import BowerLogger = require('bower-logger');
 import StandardRenderer = require('bower/lib/renderers/StandardRenderer');
 import BowerProject = require('bower/lib/core/Project');
 
+const exec = util.promisify(child_process.exec);
```
rictic: This is sadly not available in node 6, we just aren't running node 6 integration tests currently (see #136)

packages/cli/src/install/install.ts
```diff
@@ -35,25 +38,37 @@ interface JsonArray extends Array<JsonValue> {}
 export interface Options {
   variants?: boolean;
   offline?: boolean;
+  npm?: boolean;
 }
 
 export async function install(options?: Options): Promise<void> {
+  if (options && options.npm) {
+    await npmInstall();
```
rictic: Equivalent and simpler: `return npmInstall()`

packages/cli/src/install/install.ts
```diff
@@ -35,25 +36,40 @@ interface JsonArray extends Array<JsonValue> {}
...
   ]);
 }
 
+async function npmInstall() {
+  logger.info('Installing npm dependencies...');
+  await new Promise((resolve, reject) => {
+    child_process.exec('npm install', {cwd: process.cwd()}, (error, stdout, stderr) => {
+      error ? reject(error) : resolve([stdout, stderr]);
+    });
+  });
```
bicknellr: @rictic, I [stole your example](https://github.com/Polymer/tools/pull/136/files#diff-10eed550dca082f37ae2a959b6f87823R29).

packages/polyserve/src/args.ts
```diff
@@ -91,8 +91,9 @@ export const args: ArgDescriptor[] = [
   },
   {
     name: 'npm',
-    description: 'Sets npm mode: component directory is "node_modules" and ' +
-        'the package name is read from package.json',
+    description: 'Sets npm mode: dependencies are installed from npm, ' +
+        'component directory is "node_modules" and the package name is read ' +
+        'from package.json',
```
bicknellr: I'm starting to think that the `--npm` flag / config option description should be made more generic because it has to match across all commands ("Sets npm mode.") and each command's help description might be a better place to put specific information about how it reacts to the flag.

packages/cli/src/install/install.ts
rictic: +1

packages/polyserve/src/args.ts
rictic: Yeah, I think we need to do some unification to the way we handle options in the CLI

PR #146, opened 2018-04-14
Don't AMD transform non-module scripts.

aomarks:
We now pay more attention to whether an HTML script has `type=module` via the HTML splitter, and only run the AMD transform on those that do.

For external scripts, we now only run the AMD transform if the file looks like a module based on whether it has import or export statements. Babel 7 can do this for us with the "unambiguous" `sourceType`.

Fixes https://github.com/Polymer/polymer-cli/issues/994


packages/build/src/js-transform.ts
```diff
@@ -175,33 +177,41 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
...
-  if (options.transformModulesToAmd) {
-    if (options.transformImportMeta === false) {
-      throw new Error(
-          'Cannot use transformModulesToAmd without transformImportMeta.');
-    }
-    doBabel = true;
-    plugins.push(...babelTransformModulesAmd);
+  if (options.transformModulesToAmd && options.transformImportMeta === false) {
+    throw new Error(
+        'Cannot use transformModulesToAmd without transformImportMeta.');
```
rictic: Can we just turn on `transformImportMeta` when `transformModulesToAmd` is specified? Like, at the initial place where we're handed `options`? Or is this a thing that indicates a bug in polymer-build?

packages/build/src/js-transform.ts
```diff
@@ -213,6 +223,17 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
...
+
+    if (options.transformModulesToAmd === true ||
+        (options.transformModulesToAmd === 'auto' &&
+         ast.program.sourceType === 'module')) {
+      doBabelTransform = true;
+      plugins.push(...babelTransformModulesAmd);
+    }
+
+    if (doBabelTransform) {
+      js = babelCore.transformFromAst(ast, js, {presets, plugins}).code!;
```
rictic: `code!` looks suspicious here. The library is telling us that it could be null, and we're overriding it. If the library is right, we're kicking the exception down the road to a less clear place.

WDYT about checking and throwing instead?

packages/build/src/js-transform.ts
```diff
@@ -175,33 +177,41 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
...
 
-  if (doBabel) {
+  if (doBabelTransform || options.transformModulesToAmd) {
+    let ast;
     try {
-      js = babelCore.transform(js, {presets, plugins}).code!;
+      ast = babylon.parse(js, {
+        // TODO(aomarks) Remove any when typings are updated for babylon 7.
+        sourceType: options.transformModulesToAmd === 'auto' ?
+            'unambiguous' as any :
```
rictic: Neat! We should do this trick in the analyzer too, to avoid double-parsing module files.

packages/build/src/js-transform.ts
aomarks: Done. I doubt it can happen without Babel itself throwing, probably just a weakness in the Babel typings (the thing it returns is used by a bunch of functions and has optional everything https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/babel-core/index.d.ts#L172), but I can't say for sure, so added a check and throw.

packages/build/src/js-transform.ts
aomarks: Done. We now always turn on the import.meta transform when the AMD transform happens, regardless of the `transformImportMeta` option. Previously it would be auto-set only if you left it undefined, but now you can set it to false explicitly and we will override. Previously we threw because it's non-sensical to ask for AMD but not import.meta (but note that the inverse is fine), but maybe just overriding is more reasonable.

I also tried to make the code around this option more clear. I'm not sure if I succeeded.

packages/build/src/js-transform.ts
```diff
@@ -224,19 +215,44 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
...
+        relativeURL = relativeURL.replace(/\\/g, '/');
+      }
+      plugins.push(rewriteImportMeta(relativeURL));
+    }
+
     if (doBabelTransform) {
-      js = babelCore.transformFromAst(ast, js, {presets, plugins}).code!;
+      const result = babelCore.transformFromAst(ast, js, {presets, plugins});
+      if (result.code === undefined) {
+        throw new Error('Unexpected undefined code result from Babel.');
```
rictic: Maybe: `Babel transform failed: resulting code was undefined`

aomarks: Done.

PR #150, opened 2018-04-15
`Feature#astNode` is now an AstNodeWithLanguage

rictic:
This removes the largest remaining use of `any` in the analyzer's API. What previously was an AST node is now wrapped with a union discriminated by the language name. This way, a user can process an ast node that might be a parse5 node, or a babel node, or even a future language, all in a type safe way.

The main place where we'll get errors without Typescript warning is is where we're doing `babel.isSpecificTypeOfNode(feature.astNode)`. The babel-types API is very permissive in the types of the values it accepts, meaning that this will not warn, but it will now always return false. We need to change that to `feature.astNode.node`.

### DIFF
--- a/packages/analyzer/src/javascript/class-scanner.ts
+++ b/packages/analyzer/src/javascript/class-scanner.ts
@@ -118,7 +118,8 @@ export class ClassScanner implements JavaScriptScanner {
     const classMap = new Map<string, ScannedClass>();
 
     for (const class_ of classFinder.classes) {
-      if (mixinClassExpressions.has(class_.astNode)) {
+      if (class_.astNode.language === 'js' &&
+          mixinClassExpressions.has(class_.astNode.node)) {
         // This class is a mixin and has already been handled as such.
         continue;
       }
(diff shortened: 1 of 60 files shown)


rictic:
I stubbed my toe on this `any` pretty hard in some modulizer work, so I thought I'd quickly go fix it up. Took slightly longer than intended, but this should remove a major piece of type unsafeness from the API.

packages/analyzer/src/javascript/class-scanner.ts
```diff
@@ -541,7 +548,8 @@ class PrototypeMemberFinder implements Visitor {
       description,
       sourceRange: this._document.sourceRangeForNode(node)!,
       warnings: [],
-      astNode: node,
+      astNode:
+          {language: 'js' as 'js', containingDocument: this._document, node},
```
TimvdLippe: Why is the cast to `as 'js'` necessary here?

rictic: It's not clear that this is an `AstNodeWithLanguage`, and TypeScript's type inference doesn't want to infer a too-specific type. For example, should the function `() => false` have the type `() => boolean` or `() => false`? TypeScript will default to `() => boolean` unless there's a signal from the author that the more specific type is preferred. One way to do that is to write `() => false as false`, another is to assign it directly to a value which expects the type `() => false`.

TimvdLippe: Ah I see. Thanks for the explanation!

rictic:
Thanks for review! I'll add Al then too as a second pass.

PR #151, opened 2018-04-16
Add `polymer lint` to the Polymer 3 element template integration test.

bicknellr:
This was specifically excluded [in #115](https://github.com/Polymer/tools/pull/115/files#diff-10eed550dca082f37ae2a959b6f87823R49) because `polymer lint` did not have a set of rules for Polymer 3 at the time.


rictic:
Nice!

PR #152, opened 2018-04-16
Separate Babel helpers from RequireJS

### LINKED ISSUE #147: AMD babel helpers clobber the main helpers
Since #141 we now emit the `interopRequireDefault` and `interopRequireWildcard` Babel helpers into our AMD loader script. However, it turns out this clobbers the main Babel helpers script (which will be present if ES5 compilation was also enabled), since both scripts define `window.babelHelpers`.

aomarks:
In #141 I made a change whereby the two Babel helpers needed for the AMD transform were added to our require.js AMD loader script (since they go hand-in-hand). However, that meant in the (rather common) case that you both ES5 compiled and AMD transformed, our require.js script would _clobber_ the main AMD helpers script. This is because the helper code generated by AMD always overwrites `window.babelHelpers`, rather than augmenting it.

In this PR I instead introduce two different Babel helper scripts: "full" and "amd". "full" includes the AMD helpers along with everything else, while "amd" only contains the AMD helpers.

This is all kind of gross and I want to vastly simplify it via https://github.com/Polymer/tools/issues/128, but this solution should unblock us in the short term.

Fixes #147


packages/polyserve/test/bower_components/test-modules/golden/test-suite-no-wct.html
```diff
@@ -1,5 +1,5 @@
...
..." at "+n,t,[i]))}},isBrowser&&!cfg.skipDataMain&&eachReverse(scripts(),function(e){if(head||(head=e.parentNode),dataMain=e.getAttribute("data-main"),dataMain)return mainScript=dataMain,cfg.baseUrl||-1!==mainScript.indexOf("!")||(src=mainScript.split("/"),mainScript=src.pop(),subPath=src.length?src.join("/")+"/":"./",cfg.baseUrl=subPath),mainScript=mainScript.replace(jsSuffixRegExp,""),req.jsExtRegExp.test(mainScript)&&(mainScript=dataMain),cfg.deps=cfg.deps?cfg.deps.concat(mainScript):[mainScript],!0}),define=function(e,r,i){var n,t;"string"!==typeof e&&(i=r,r=e,e=null),isArray(r)||(i=r,r=null),!r&&isFunction(i)&&(r=[],i.length&&(i.toString().replace(commentRegExp,commentReplace).replace(cjsRequireRegExp,function(e,i){r.push(i)}),r=(1===i.length?["require"]:["require","exports","module"]).concat(r))),useInteractive&&(n=currentlyAddingScript||getInteractiveScript(),n&&(!e&&(e=n.getAttribute("data-requiremodule")),t=contexts[n.getAttribute("data-requirecontext")])),t?(t.defQueue.push([e,r,i]),t.defQueueMap[e]=!0):globalDefQueue.push([e,r,i])},define.amd={jQuery:!0},req.exec=function(text){return eval(text)},req(cfg)}})(this,"undefined"===typeof setTimeout?void 0:setTimeout);</script>
```
aomarks: @rictic Again, sorry for the horrible test output. What's happening here in the Polyserve output is that the two AMD Babel helpers are no longer part of the requirejs script. This is actually another reason my earlier approach wasn't a great idea -- in Polyserve we do inline Babel helpers, so there was no reason to include it in this static script.

packages/build/src/html-transform.ts
```diff
@@ -51,10 +51,14 @@ export interface HtmlTransformOptions {
...
-   * document.
+   * needed when this is the entry point HTML document and ES5 compilation or
+   * AMD transform is enabled.
+   *
+   * If "full", includes the helpers needed for both ES5 compilation and the AMD
+   * transform. If "amd", includes only the helpers needed for the AMD
+   * transform.
    */
-  injectBabelHelpers?: boolean;
+  injectBabelHelpers?: false|'full'|'amd';
```
rictic: What about `'none'|'full'|'amd'` instead?

packages/build/src/html-transform.ts
```diff
@@ -152,8 +156,21 @@ export function htmlTransform(
...
+    let helpers;
+    switch (options.injectBabelHelpers) {
+      case 'full':
+        helpers = getMinfiedBabelHelpersFull();
+        break;
+      case 'amd':
+        helpers = getMinifiedBabelHelpersAmd();
+        break;
+      default:
+        throw new Error(
```
rictic: Use the type checker for completeness checking:

```typescript
const never: never = options.injectBabelHelpers;
throw new Error(`Unknown injectBabelHelpers value: ${never}`);
```

packages/build/src/html-transform.ts
aomarks: Done.

packages/build/src/html-transform.ts
aomarks: Done.

PR #153, opened 2018-04-17
Don't build when executing test:unit or test:integration

usergenic:
Install steps already run build tasks and all package 'test' scripts also run build tasks, so travis is spending a lot of time unnecessarily running build tasks.  Attempt to solve this by adding `test:unit` npm scripts to all packages and explicitly run those scripts instead.

Also removed the build gulp task dependencies on test:unit or test:integration where found.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,10 +12,11 @@
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "build": "lerna run build",
-    "test": "lerna run test",
-    "test:windows": "lerna run test --ignore polyserve",
+    "test": "lerna run build && lerna run test:unit",
     "test:integration": "lerna run test:integration",
-    "test:integration:windows": "lerna run test:integration"
+    "test:integration:windows": "lerna run test:integration",
+    "test:unit": "lerna run test:unit",
+    "test:unit:windows": "lerna run test:unit --ignore polyserve"
   },
   "devDependencies": {
     "lerna": "^2.10.1"
--- a/packages/build/gulpfile.js
+++ b/packages/build/gulpfile.js
@@ -52,7 +52,9 @@ gulp.task('compile', () => {
       .pipe(gulp.dest('lib'));
 });
 
-gulp.task('test', ['build'], function() {
+gulp.task('test', ['build', 'test:unit']);
+
+gulp.task('test:unit', function() {
   return gulp.src('lib/test/**/*_test.js', {read: false}).pipe(mocha({
     ui: 'tdd',
     reporter: 'spec',
--- a/packages/editor-service/package.json
+++ b/packages/editor-service/package.json
@@ -17,8 +17,9 @@
     "build": "tsc",
     "prepublishOnly": "npm run clean && npm run build",
     "lint": "tslint --project ./tsconfig.json",
-    "test": "npm run clean && npm run build && mocha && npm run lint",
-    "test:watch": "tsc-then -- mocha",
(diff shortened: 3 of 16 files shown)


packages/editor-service/package.json
```diff
@@ -17,8 +17,9 @@
     "build": "tsc",
     "prepublishOnly": "npm run clean && npm run build",
     "lint": "tslint --project ./tsconfig.json",
-    "test": "npm run clean && npm run build && mocha && npm run lint",
-    "test:watch": "tsc-then -- mocha",
+    "test": "npm run clean && npm run build && npm run test:unit && npm run lint",
+    "test:unit": "mocha",
+    "test:watch": "tsc-then -- npm run test:unit",
```
rictic: running `mocha` directly is generally better in test:watch, because if there's a failure when doing `npm run foo` then npm outputs like 20-30 lines about how the failure isn't npm's fault

packages/linter/package.json
```diff
@@ -46,8 +46,9 @@
     "clean": "touch lib && rm -rf lib && mkdir lib",
     "format": "find src | grep \"\\.[jt]s$\" | xargs clang-format --style=file -i",
     "lint": "tslint -c tslint.json src/*.ts src/**/*.ts",
-    "test": "npm run build && mocha && npm run lint",
-    "test:watch": "tsc-then -- mocha",
+    "test": "npm run build && npm run test:unit && npm run lint",
+    "test:unit": "mocha",
+    "test:watch": "tsc-then -- npm run test:unit",
```
rictic: prefer just tsc-then -- mocha

packages/build/gulpfile.js
```diff
@@ -52,7 +52,9 @@ gulp.task('compile', () => {
       .pipe(gulp.dest('lib'));
 });
 
-gulp.task('test', ['build'], function() {
+gulp.task('test', ['build', 'test:unit']);
+
+gulp.task('test:unit', function() {
```
rictic: Looks like build is missing a corresponding change to its package.json?

package.json
```diff
@@ -11,11 +11,12 @@
...
-    "test": "lerna run test",
-    "test:windows": "lerna run test --ignore polyserve",
-    "test:integration": "lerna run test:integration",
-    "test:integration:windows": "lerna run test:integration"
+    "build": "lerna run build --stream",
+    "test": "npm run build && npm run test:unit",
+    "test:integration": "lerna run test:integration --stream",
+    "test:integration:windows": "lerna run test:integration --stream",
+    "test:unit": "lerna run test:unit --stream",
+    "test:unit:windows": "lerna run test:unit --stream --ignore polyserve"
```
rictic: +1 to --stream, nice catch

usergenic:
Note: I put the `--stream` flag on the lerna commands primarily because of integration tests.  When you run any task with lerna that exceeds 10 minutes, that means no output is generated for those 10 minutes and travis just gives up.  Using `--stream` ensures output is generated in realtime instead of after the entire task is completed, which makes travis happier.  Its also just nice to get feedback that sh*t is happening.

packages/editor-service/package.json
usergenic: ok.

packages/build/gulpfile.js
usergenic: Good catch.  Added

packages/linter/package.json
usergenic: I've changed all `npm run test:unit` inside of npm scripts to `mocha` where appropriate.

PR #154, opened 2018-04-17
Updates CLI to use '@octokit/rest' instead of 'github'.

bicknellr:


packages/cli/src/github/github.ts
```diff
@@ -156,21 +154,22 @@ export class Github {
...
-    const releases: GitHubApi.Release[] = await this._github.repos.getReleases({
+    const response: Octokit.AnyResponse = await this._github.repos.getReleases({
       owner: this._owner,
       repo: this._repo,
       per_page: 100,
     });
+    const releases = response.data;
     const validReleaseVersions =
-        releases.filter((r) => semver.valid(r.tag_name)).map((r) => r.tag_name);
+        releases.filter((r: any) => semver.valid(r.tag_name)).map((r: any) => r.tag_name);
```
bicknellr: Unfortunately, it seems like their `index.d.ts` no longer has types for responses.

rictic:
What's the advantage of `octokit` over `github`?

TimvdLippe:
There is no advantage other than that the old package name is no longer updated. They simply renamed their package and continued development under the new name.

bicknellr:
I was trying to figure out how to retrieve a branch instead of a release and noticed they had updated and couldn't find documentation for the version we were using.

TimvdLippe:
See also https://github.com/octokit/rest.js/releases/tag/v14.0.0 btw

packages/cli/src/github/github.ts
```diff
@@ -156,21 +154,29 @@ export class Github {
    * Get all Github releases and match their tag names against the given semver
    * range. Return the release with the latest possible match.
    */
-  async getSemverRelease(semverRange: string): Promise<GitHubApi.Release> {
+  async getSemverRelease(semverRange: string): Promise<any> {
```
rictic: Switching from types to `any` is a pretty big regression. I'd rather not update the library if that's the primary change.

rictic: If there's an expectation that the newer version of the library is better then I'm on board

TimvdLippe: I think this is blocked on https://github.com/octokit/rest.js/pull/732

PR #157, opened 2018-04-17
Import WCT

usergenic:
Import web-component-tester from https://github.com/Polymer/web-component-tester into `./packages/web-component-tester`.

Other than the import, this involved making a `test:unit` and `test:integration` npm script for `wct`, updating `.travis.tml` and adding a missing `@types` item to `./packages/cli/package.json` because of stricter typescript compilation rules in `Polymer/tools`.  Also had to add `@types/resolve` and `@types/spdy` to wct's package.json as well as make wct-local and wct-sauce actual devDependencies since tests depend on them.

Also fixed accidental parallel gulp tasks resulting from prior adjustments to gulpfiles I did without realizing they would run in parallel.


package.json
```diff
@@ -14,9 +14,9 @@
     "build": "lerna run build --stream",
     "test": "npm run build && npm run test:unit",
     "test:integration": "lerna run test:integration --stream",
-    "test:integration:windows": "lerna run test:integration --stream",
+    "test:integration:windows": "lerna run test:integration --stream --ignore polymer-linter --ignore web-component-tester",
```
aomarks: linter's integration test script was removed, so we don't need this `--ignore polymer-linter`

usergenic: Cool.   Removed!

PR #159, opened 2018-04-17
Fix broken test:unit gulp task.

aomarks:
gulp dependencies run in parallel, so this recent change broke the test script.

### DIFF
--- a/packages/build/gulpfile.js
+++ b/packages/build/gulpfile.js
@@ -57,7 +57,7 @@ gulp.task('compile', () => {
       .pipe(gulp.dest('lib'));
 });
 
-gulp.task('test', ['build', 'test:unit']);
+gulp.task('test', (done) => runSeq('build', 'test:unit', done));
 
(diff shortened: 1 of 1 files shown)


usergenic:
gulp dependencies run in parallel?!?!

🤯

aomarks:
> Yikes! I had no idea they were parallelized. Nice find. There may be others... Will keep my eyes peeled.

Yeah: https://github.com/gulpjs/gulp/blob/v3.9.1/docs/API.md#gulptaskname--deps-fn

Looks like the new API (still @next) makes this much more clear by introducing `.parallel` and `.series`: https://github.com/gulpjs/gulp/blob/master/docs/API.md

usergenic:
I needed to fix this one and a few others in my https://github.com/Polymer/tools/pull/157 branch as well.  Commit: https://github.com/Polymer/tools/pull/157/commits/4f615eb43a749dcde720dde2a1823b7ec8c7a671

PR #160, opened 2018-04-17
[analyzer] support all default param types

### LINKED ISSUE #132: [analyzer] Method's default parameter values are included in the parameter name
Analyzer includes the default value as part of the parameter name. This causes the JSDoc comments to be ignored as the names do not match.

Input:
``` javascript
...

/**
   * My first method
   * @param {object} param My first parameter
   */
myMethod (param = { foo: 'bar' }) {
  ...
}
```

Output:
``` json
...
  "params": [
    {
      "name": "param = { foo: 'bar' }"
    }
  ]
...
```

Expected Output:
``` json
...
  "params": [
    {
      "name": "param",
      "type": "object",
      "description": "My first parameter",
      "?defaultValue": "{ foo: 'bar' }"
    }
  ]
...
```

43081j:
Fixes #132.

We were only accepting literal right hand values before, but really i suppose we don't care what it is. it gets turned into a string anyway.


rictic:
`master` was broken when you sent this over, mind rebasing real quick?

rictic:
Pushed up a merge commit, if it passes I'll merge this in

43081j:
thanks for that, had gone mobile yesterday 🙊 

the tests failed because of an ordering mistake in the assertions by the looks of it. will fix it shortly

rictic:
LGTM!

PR #164, opened 2018-04-17
ParsedDocument#astNode was `any`.

rictic:


packages/analyzer/src/parser/document.ts
```diff
@@ -183,7 +185,7 @@ export class UnparsableParsedDocument extends ParsedDocument {
   type: string = 'unparsable';
   constructor(url: ResolvedUrl, contents: string) {
     super({
-      ast: null,
+      ast: null as any,
```
aomarks: why is this safe?

rictic: Turns out it wasn't necessary. Removed.

PR #166, opened 2018-04-18
Use a single shared top-level tslint

rictic:
Also run `tslint -p --fix` across all of the repos.


rictic:
It doesn't change packages/common yet.

PR #168, opened 2018-04-18
Added a polyserve option doNotCompile and speed up WCT tests

usergenic:
`--do-not-compile`/ `doNotCompile` = Array of glob string patterns to match against request url paths in polyserve to skip babelCompile middleware.  WCT uses this to specify NOT to compile the `/components/web-component-tester/browser.js` file to speed tests up.


packages/polyserve/src/compile-middleware.ts
```diff
@@ -69,6 +71,11 @@ export function babelCompile(
       if (isPolyfill.test(request.url)) {
         return false;
       }
+      if (doNotCompile.length > 0 &&
+          !!doNotCompile.find((p) => minimatch(request.url, p))) {
+            console.log('SKIPPY MA-DOO-DAH',request.url);
```
aomarks: lol

packages/polyserve/src/args.ts
```diff
@@ -36,6 +36,14 @@ export const args: ArgDescriptor[] = [
     type: String,
     defaultValue: 'auto',
   },
+  {
+    name: 'do-not-compile',
+    description: 'Disables compilation for file(s) matching given pattern. ' +
```
aomarks: Can we clarify what we mean by compilation? As currently written, it bypasses ES5 compilation, but also name-to-path transformation and AMD transformation. compile-middleware is already probably too aggressive in this regard, since the `compile=never` returns `false` from `shouldTransform` in the same way.

packages/polyserve/src/args.ts
```diff
@@ -36,6 +36,14 @@ export const args: ArgDescriptor[] = [
     type: String,
     defaultValue: 'auto',
   },
+  {
+    name: 'do-not-compile',
```
aomarks: Now we have `--compile` and `--do-not-compile`, which might be slightly confusing. I guess we should just document the interaction between the two, especially that this takes precedence over `--compile=always`?

(I was also thinking we could update the `--compile` flag to take glob patterns (and minimatch already supports the `!` operator), but then you still have the `auto` vs `always` difference that needs to be represented somehow. Not sure if this idea makes sense.)

usergenic: Possible alternate names:

- `--pass-thru`
- `--do-not-modify`
- `--serve-as-is`
- `--as-is`
- `--no-transform`
- `--no-rewrite`
- `--immutable`
- `--static`
- `--constant`
- `--invariant`

aomarks: But do we *want* to disable all the transforms, or is this just about ES5 compilation?

aomarks:
Are we still going to do this? Can close PR otherwise?

usergenic:
Closing PR because not important right now and this is rotting on the vine.

PR #169, opened 2018-04-18
[cli] Adds an option to init to use the Polymer 3.x compatible PSK prerelease.

### LINKED ISSUE #162: [cli] init: Add an option to init from the Polymer 3 compatible PSK prerelease.
Here's the Polymer 2 version:

https://github.com/Polymer/tools/blob/67f781d368272f811607c92e74bbf8ad4e75f63f/packages/cli/src/init/init.ts#L62-L71

Once there's a prerelease tagged for a Polymer 3 compatible PSK, this should mostly be a copy-paste.

bicknellr:
Fixes #162.


packages/cli/src/init/github.ts
```diff
@@ -76,9 +76,7 @@ export function createGithubGenerator(githubOptions: GithubGeneratorOptions):
     }
 
     install() {
-      this.installDependencies({
-        npm: false,
```
rictic: Did we have `npm: false` for just performance reasons? I'm wondering if this will cause issues for our existing generators. I'm kinda assuming no, it'll just be a bit slower?

I guess the only case I can think of would be if yeoman failed if we don't say `npm: false` but there also isn't a package.json file.

bicknellr: I'm not sure why this was here - yeoman seems to figure this out on its own. When I init a Polymer 2 template with this change it specifically prints `Running bower install for you to ...` and when I init one of the Polymer 3 templates it prints `Running npm install for you to ...`.

bicknellr: Also, each of the built-in templates only has one of `bower.json` or `package.json`, PSK for Polymer 3 is soon to only have `package.json`, and Shop for Polymer 2 only has `bower.json`.

Issue #170, opened 2018-04-18
The definition finder editor-service tests are a bit flaky


rictic:
As a first pass I think we should skip them for the time being.

PR #172, opened 2018-04-18
Skip DefinitionFinder tests for the moment.

rictic:
See https://github.com/Polymer/tools/issues/170


rictic:
Merging before tests because this is obviously safe

PR #171, opened 2018-04-18
Run tests in parallel, explicitly.

usergenic:
This is more significant for integration tests, which seem to not be running in parallel right now for some reason.

PR #174, opened 2018-04-18
Capture and assert on error output.

rictic:
I noticed when trying out #171 that `build` emits a super scary looking stack trace as part of its expected output when testing. This captures and asserts on that output instead.


rictic:
I'd like to get us to a point where our tests emit no logs themselves (when they pass)

PR #184, opened 2018-04-19
Try caching node_modules and packages/*/node_modules on travis.

usergenic:
I don't know if travis supports `*` in the directories list for caching, but I'm trying that now.


usergenic:
This optimization is still premature, closing for now.

PR #185, opened 2018-04-19
Limit travis and appveyor to building master branch and PR commits only.

usergenic:
Decrease travis utilization by 50%


rictic:
2x reduction in utilization!

PR #188, opened 2018-04-19
Added custom typings for the inplace method of stable.d.ts to avoid implicit any warnings.

usergenic:
This was causing warnings to emit on build.  Just getting rid of that.


usergenic:
Note to @rictic I decided instead of adding explicit type to the anon function, I added custom typing for the stable package `inplace` method used.

bicknellr:
We need to fix master: this PR caused tests to start failing which are not related to the recent WCT errors.

PR #189, opened 2018-04-19
Disable minify-constant-folding plugin.

aomarks:
Disable Babel `minify-constant-folding` plugin when minifying. This
plugin has a bug that breaks when a constant is exported from a module
(https://github.com/babel/minify/issues/820).

Fixes https://github.com/Polymer/polymer-cli/issues/998

### DIFF
--- a/packages/build/src/js-transform.ts
+++ b/packages/build/src/js-transform.ts
@@ -69,8 +69,15 @@ const babelSyntaxPlugins = [
   require('@babel/plugin-syntax-import-meta'),
 ];
 
-const babelPresetMinify =
-    require('babel-preset-minify')({}, {simplifyComparisons: false});
+const babelPresetMinify = require('babel-preset-minify')({}, {
+  // Disable the minify-constant-folding plugin because it has a bug relating to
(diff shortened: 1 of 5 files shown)


packages/build/src/js-transform.ts
```diff
@@ -69,8 +69,15 @@ const babelSyntaxPlugins = [
...
 
-const babelPresetMinify =
-    require('babel-preset-minify')({}, {simplifyComparisons: false});
+const babelPresetMinify = require('babel-preset-minify')({}, {
+  // Disable the minify-constant-folding plugin because it has a bug relating to
+  // invalid substitution of constant values into export specifiers:
+  // https://github.com/babel/minify/issues/820
+  evaluate: false,
+
+  // TODO(aomarks) Find out why we disabled this plugin.
```
rictic: s/plugin/option/?

aomarks: The options map to plugins (https://github.com/babel/minify/tree/master/packages/babel-preset-minify#options)

PR #190, opened 2018-04-20
Add --auto-base-path flag to build

### LINKED ISSUE #175: [cli] Add --auto-base-path flag to build
Basically I want the feature implemented in https://github.com/Polymer/polymer-cli/pull/920. With this I can have two npm scripts:
- `build:prpl-server` which uses this flag
- `build:static` which uses the same polymer.json to make several build outputs, each of which I can test with a static web server

keanulee:
Fixes #175. Code changes previously reviewed (but unmerged) in https://github.com/Polymer/polymer-cli/pull/920

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -31,6 +31,10 @@ import {Command, CommandOptions} from './command';
 
 const logger = logging.getLogger('cli.command.build');
 
+// This flag is special because it doesn't necessarily trigger a one-off single
+// build.
+const autoBasePath = 'auto-base-path';
+
 export class BuildCommand implements Command {
   name = 'build';
   aliases = [];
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -133,6 +143,7 @@ export class BuildCommand implements Command {
       ProjectBuildOptions {
     const buildOptions: ProjectBuildOptions = {};
     const validBuildOptions = new Set(this.args.map(({name}) => name));
+    validBuildOptions.delete(autoBasePath);  // This flag is special.
```
aomarks: `// This flag is special. It corresponds to the top-level config, not an individual build.`

abdonrd:
For reference, some related work here: https://github.com/Polymer/polymer-project-config/pull/48

PR #193, opened 2018-04-20
Add retries for all the tests

usergenic:
This is mostly just adding `retries` of 3 to mocha configs.  The only place I had to do something different was in the WCT integration tests `browser.ts` basically does a lot of work in the `before()` and that isn't covered by retries, so I manually set up a try/catch block in a while loop that retries three times on uncaught exception.

### DIFF
--- a/packages/analyzer/gulpfile.js
+++ b/packages/analyzer/gulpfile.js
@@ -107,7 +107,6 @@ task('test', ['build'], () => {
   return gulp.src('test/**/*_test.js').pipe(mocha({
     ui: 'tdd',
     reporter: 'spec',
-    retries: 3,
   }));
 });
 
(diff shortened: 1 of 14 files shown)


rictic:
`retries` is a code smell for me, as it can be hiding real issues, and it makes it harder to read the output. 

I'd prefer that we add it only where we know we need it, so wct and the editor service.

usergenic:
Well, ok. I'll limit to those 2 for now.  Still rather just have retries across the board and figure how to show retry counts in the reporters but I'll make the change

usergenic:
Also, if I get occasional failures in another suite, I'm adding retries back in for that suite. ;-)

usergenic:
Watching travis *intensifies*

rictic:
Merging this to get the retry goodness asap!

PR #194, opened 2018-04-20
Run integration tests in parallel on travis.

usergenic:
Before...
screen shot 2018-04-20 at 3 24 11 pm

After...
screen shot 2018-04-20 at 5 35 18 pm


usergenic:
Ran into a flaky CLI integration test, possibly exacerbated by parallel run with WCT, but easily addressed with mocha retries, so added mocha retries just for CLI integration tests.
screen shot 2018-04-20 at 3 44 11 pm

PR #195, opened 2018-04-20
Rewritten dynamic `import()` calls now test if a bundle has exports b…

keanulee:
…efore attempting to destructure bundle

Fixes https://github.com/Polymer/polymer-cli/issues/995

### DIFF
--- a/packages/bundler/src/es6-rewriter.ts
+++ b/packages/bundler/src/es6-rewriter.ts
@@ -250,7 +250,7 @@ export class Es6Rewriter {
    *
    * After:
    *     import('./bundle_1.js')
-   *         .then(bundle => bundle && bundle.$moduleA)
+   *         .then(bundle => bundle && bundle.$moduleA || {})
    *         .then((moduleA) => moduleA.doSomething());
    */
   private _rewriteDynamicImport(
@@ -309,11 +309,14 @@ export class Es6Rewriter {
         [babel.arrowFunctionExpression(
             [babel.identifier('bundle')],
             babel.logicalExpression(
-                '&&',
-                babel.identifier('bundle'),
-                babel.memberExpression(
+                '||',
+                babel.logicalExpression(
+                    '&&',
                     babel.identifier('bundle'),
-                    babel.identifier(exportName))))]);
+                    babel.memberExpression(
+                        babel.identifier('bundle'),
+                        babel.identifier(exportName))),
+                babel.objectExpression([])))]);
     rewriteObject(importCallExpression, thenifiedCallExpression);
   }
 
--- a/packages/bundler/src/es6-rewriter.ts
+++ b/packages/bundler/src/es6-rewriter.ts
@@ -250,7 +250,7 @@ export class Es6Rewriter {
    *
    * After:
    *     import('./bundle_1.js')
-   *         .then(({$moduleA}) => $moduleA)
+   *         .then(bundle => bundle && bundle.$moduleA)
    *         .then((moduleA) => moduleA.doSomething());
    */
   private _rewriteDynamicImport(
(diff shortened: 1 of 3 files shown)


packages/bundler/src/es6-rewriter.ts
```diff
@@ -304,19 +304,18 @@ export class Es6Rewriter {
...
-              babel.objectPattern(
-                  [babel.objectProperty(
-                       babel.identifier(exportName),
-                       babel.identifier(exportName),
-                       undefined,
-                       // tslint:disable-next-line: no-any
-                       true) as any]),
-            ],
-            babel.identifier(exportName))]);
+      babel.memberExpression(
```
aomarks: can you run `npm format`

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
```diff
@@ -199,9 +199,7 @@ suite('Es6 Module Bundling', () => {
           await bundler.bundle(await bundler.generateManifest([aUrl]));
       assert.deepEqual(documents.get(aUrl)!.content, heredoc`
         async function go() {
-          const b = await import('./b.js').then(({
-            $b
-          }) => $b);
+          const b = await import('./b.js').then(bundle => bundle && bundle.$b);
```
rictic: Should this be: `bundle => bundle && bundle.$b || {}`?

Because it's totally legal to do:

```javascript
const b = await import('./b.js');
b.thisPropertyDoesNotExist === undefined
```

packages/bundler/src/es6-rewriter.ts
keanulee: VS code formatted this file on save - pretty much the whole file was modified, but I just committed these lines.

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
rictic: I mean ideally AMD would make the default export `{}` so that `$b` would be {} so long as `./b.js` exists, regardless of whether it exports anything

keanulee: So without bundling, `import('./foo.js')` will resolve to an empty `Module` object (which you could attempt to get an undefined property from), but the AMD transform is a Promise that resolves to `undefined` (which you can't destructure). By giving you `undefined` when importing bundles, this is consistent with the AMD transform.

Another possible fix would be to make sure AMD transform always gives you an empty object, but that would need to be done in [the Babel transform](https://github.com/babel/babel/blob/a4795408b47e713aa2635d54197343be77e30f35/packages/babel-plugin-transform-modules-amd/src/index.js#L46) AFAICT

keanulee: To be clear, that wouldn't work in the unbundled, AMD transform case either (e.g. at dev time with `polymer serve` on a non-ES modules browser like Firefox) - if a developer forgets to export something, the destructure call will still error. That is, no guarantees are provided by us if you try to use something that isn't exported.

keanulee:
In the interest of getting this blocker PR merged I made the suggested change, though it probably should be updated once the AMD transform is guaranteed to export an object (i.e. `bundle` is definitely truth-y).

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
usergenic: @keanulee after reviewing, the bundle-producing code is probably not worth changing as it seems correct.  I don't exactly mind having a safety in here like you've done to guard against errors introduced by bundler's dynamic import statement transform not playing nice with the AMD transform, but this ultimately feels like a problem with that transform.

Also there's a subtle issue with respect to what you said...
>  if a developer forgets to export something, the destructure call will still error

...because it isn't actually an error to destructure a property from a module object if that name  doesn't exist.  It's an error to try and import a non-exported name:
```
import {notAThing} from 'something'; // error
```
...but it is okay to destructure a value from a module object that is not exported, which is functionally what a dynamic import case is like:
```
import * as stuff from 'something';
const {notAThing} = stuff; // not an error
```
However, if we make a specific case for `|| {}` in the dynamic import `.then` code we are maybe being inconsistent, since a non-dynamic import of `$b` as in the following would be (at best?) an undefined value after AMD transform instead of an empty object:
```
// originally: import * as stuff from './b.js';
import {$b as stuff} from './bundle.js';
stuff.value // error in AMD transform output when value not exported
```

So with all that reasoning... I will approve this PR since it addresses the most common problem which is dynamically imported fragment that doesn't export anything blowing up after bundling.  The next step is addressing the behavior of the AMD transform so it doesn't cause downstream problems for transforming modules without exports.

usergenic:
@keanulee I just looked at Travis output and the failures displayed are transient.  You can squash + merge this now.

PR #196, opened 2018-04-20
Include `.gitignore`s from templates when publishing.

bicknellr:


bicknellr:
Yeah, thanks for the help, I had no idea there was an `.npmignore`! cc @katejeffreys, this is the bug you found earlier today.

katejeffreys:
\o/

bicknellr:
(restarted two builds that timed out)

Issue #206, opened 2018-04-23
[serve] Resources are incorrectly cached when restarting polyserve with different transform options


rictic:
Repro:
  - `polymer serve --compile=always`
  - load file in browser, code is compiled, etag and last-modified-time headers are sent
  - kill serve and restart with `polymer serve --compile=never`
  - reload file in browser. get a 304 because of etag and modified time, so you see compiled results when running with `--compile=never`

PR #207, opened 2018-04-23
Disable cache headers that do not account for transforms.

rictic:
Fixes https://github.com/Polymer/tools/issues/206


rictic:
updated changelog

rictic:
Tests failing due to https://github.com/Polymer/tools/pull/207

aomarks:
> Tests failing due to #207

Added merge commit. Might pass.

PR #209, opened 2018-04-23
Don't wait for *all* unit tests to complete before running integration tests.

usergenic:
This might shave a couple of minutes off of the travis runs to start CLI integration tests immediately after CLI unit tests complete instead of waiting for all package unit tests to complete.


.travis.yml
```diff
@@ -12,6 +12,8 @@ env:
   global:
   - secure: IbrybaUaJDiwNh8VodFfTIAdSw2Nl4DGRh2oDIO2vL8SPgGilwxY0go5DWrxAG0aP+YIXFkK+SpblFppQqqxnAoLVYmd7kFSJoOZraNqmcJqUYSzrHrWArmuiGzY8gMgOWTsnbzVVNRldN8tpoSwpQB/+YhCtbYb9eqvw+u5Izk=
   - secure: grvHX6c0ReQwRsuQZferI+uyGsKlubujURKuJB8coNdA9lgZcvf+nmCkNBpSML83hSjQkJUTt8XnucNNgL/1VxUhgy0O/wkTUEmcDAU1b80As8NJkbreJcc3HFASA68wou4HoJsFmJV2cpBYoAJJSX+HyT6mYd6pBh09tPfOeBg=
+  matrix:
+  - TEST_COMMAND="run test:all:parallel"
```
rictic: is this matrix part necessary? why not just inline it in the script section?

usergenic: - One thing is, I missed seeing a definition of the command run in the travis UI.
- Second thing is, I updated the test:all:parallel npm script to fix the issue and then totally didn't realize the .travis.yml didn't even use that script, and I might have noticed it if the UI showed the command that was run.

rictic:
This seems reasonable, but I want to first get tests passing consistently before optimizing further.

PR #210, opened 2018-04-23
Update goldens and package locks/shrinkwrap

### LINKED ISSUE #208: [build] CLI integration tests failing, perhaps due to a change in babel helpers
https://gist.github.com/usergenic/5d0f10fca8122564faa6eb595e449cd5

aomarks:
Theory: there was a new release of many Babel packages, which changed the Babel helpers JS output a little bit, which broke our CLI build integration tests. It affected Node 6 on Travis because the package-lock is ignored in that version, and it affected AppVeyor because we explicitly delete package-locks.

Fixes #208


usergenic:
Riddle me this: why are these tests still failing?

PR #212, opened 2018-04-23
Force colors in all the mochas and make all unit and integration test npm scripts run mocha directly.

usergenic:
Our use of mocha and our mocha.opts were inconsistent and I thought it would be good to handle them consistently as we move standardize things across monorepo.  Normalizing the output will make CI output easier to grok too.


usergenic:
I removed the globs from the opts files because I couldn't find a way to override them when they were there.  This was mostly an issue when the package needed to run integration tests folders exclusively or files and unit test folders exclusively.

Maybe there's another way to solve that.  I didn't know what it was off-hand.

Note: in the npm test scripts, I provide the `"lib/test/**/*_test.js"` globs inside quotes so that mocha parses them not the shell.

rictic:
> Note: in the npm test scripts, I provide the "lib/test/**/*_test.js" globs inside quotes so that mocha parses them not the shell.

If that's the case then `lib/**/*_test.js` should be the same as `lib/*_test.js lib/**/*_test.js`

usergenic:
This PR got weirdly messy due to merges and such.  Closing and superceding with cleaner PR https://github.com/Polymer/tools/pull/267

PR #216, opened 2018-04-24
Pin @babel/plugin-transform-classes to v7.0.0-beta.35

bicknellr:
babel/babel#7020 (included in v7.0.0-beta.36) breaks our assumptions about how babel will transform ES6 classes. Particularly, it changes the way that compiled ES6 classes which extend (what babel assumes to be) native constructors such that the subclasses no longer call into the parent classes' constructors. This is allows the transformation to avoid a specific type of error when extending native constructors: most must be constructed with `new` (or `super` / `Reflect.construct`, which aren't available in `ES2015`) and will throw when called using `call` or `apply`. However, the polyfill needs the user to call into the polyfilled constructor to retrieve the exact element which is being upgraded and so that the user's constructors are called with that specific element as `this`.

This PR manually assembles the set of transforms in @babel/preset-es2015 (with our particular configuration) for the purpose of being able to pin @babel/plugin-transform-classes to v7.0.0-beta.35 .

- [x] tests

---

https://github.com/babel/babel/issues/7506 - a bug about this issue
https://github.com/babel/babel/pull/7533 - related PR to fix bug above


packages/build/src/js-transform.ts
```diff
@@ -29,8 +29,30 @@ import isWindows = require('is-windows');
 // minifying but not compiling?
 
 // Syntax and transform plugins for ES2015.
-const babelPresetEs2015NoModules =
-    [require('@babel/preset-es2015'), {modules: false}];
+const customBabelPresetEs2015 = {
+  plugins: [
+    [require("@babel/plugin-transform-template-literals"), { loose: false, spec: false }],
```
aomarks: I think all of these parameters are default false except for the regenerator ones? Could just leave off the ones that aren't changing a default.

packages/build/src/js-transform.ts
```diff
@@ -29,8 +29,30 @@ import isWindows = require('is-windows');
 // minifying but not compiling?
 
 // Syntax and transform plugins for ES2015.
```
aomarks: Add note linking to bug about class plugin and mention its why we aren't using the preset.

packages/build/src/js-transform.ts
```diff
@@ -168,7 +190,7 @@ export function jsTransform(js: string, options: JsTransformOptions): string {
   }
   if (options.compileToEs5) {
     doBabelTransform = true;
-    presets.push(babelPresetEs2015NoModules);
+    presets.push(customBabelPresetEs2015);
```
aomarks: `plugins.push(...`? Probably works either way, but seems more clear now that it's not technically a preset.

packages/build/src/js-transform.ts
bicknellr: You were right, almost all of these were the defaults.

packages/build/src/js-transform.ts
bicknellr: added

aomarks:
If there's a straightforward way to check that the bad transform is not happening, mind adding a regression test in js-transform_test?

packages/build/src/js-transform.ts
bicknellr: AFAICT, this would change the order? https://babeljs.io/docs/plugins/#pluginpreset-ordering

bicknellr: https://babeljs.io/docs/plugins/#creating-a-preset kinda makes it sound like this counts as a preset.

aomarks: I think a preset is just an array of plugins, and yes it does affect ordering (if you look above in the transform functions I mentioned how things are ordered). I guess this is fine, and it definitely won't change the ordering!

bicknellr:
I've added a test that checks for whether or not [this line](https://github.com/babel/babel/pull/7020/files#diff-ba75bd4d116cc90ad8c8002e3010e821R452) (the broken constructor) is injected into the output when creating a class that extends from `HTMLElement`.

packages/build/package.json
```diff
@@ -24,19 +24,38 @@
...
-    "@babel/plugin-external-helpers": "^7.0.0-beta.42",
-    "@babel/plugin-proposal-async-generator-functions": "^7.0.0-beta.42",
-    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.42",
-    "@babel/plugin-syntax-async-generators": "^7.0.0-beta.42",
+    "@babel/core": "^7.0.0-beta.46",
+    "@babel/plugin-external-helpers": "^7.0.0-beta.46",
+    "@babel/plugin-proposal-async-generator-functions": "^7.0.0-beta.46",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0-beta.46",
+    "@babel/plugin-syntax-async-generators": "^7.0.0-beta.46",
     "@babel/plugin-syntax-dynamic-import": "^7.0.0-beta.43",
```
bicknellr: I'm going to bump this one to 46 also, just to keep everything else at the same version.

packages/build/src/test/js-transform_test.ts
```diff
@@ -336,4 +336,21 @@ suite('jsTransform', () => {
...
+          super();
+          this.x = 1234;
+        }
+      }
+
+      window.customElements.define("test-element", TestElement);
+    `);
+    const result = jsTransform(input, {compileToEs5: true});
+
+    assert.notInclude(result, "function Wrapper() {}");
```
aomarks: What about checking for `wrapNativeSuper`, since it's the name of the helper function that contains this line? Maybe less brittle.

aomarks: I think just wrapNativeSuper is fine. If you did do both, make sure it's an *or*, since an *and* would be _more_ brittle.

bicknellr: Ok, I'll just do `wrapNativeSuper` then.

aomarks:
We should probably test with shop or something, just to make sure it actually works at runtime. Like I could imagine a breakage due to the helpers and the transformer being out of sync.

bicknellr:
I've been trying with a `polymer init`-created PSK, manually modified to force the custom elements polyfill and w/ UA spoofed to IE11. I *think* it's working as expected but I'm going to ask @keanulee to double check for me since he has better knowledge of the conditions that lead to this problem.

bicknellr:
@keanulee thanks! I was just about to walk over and ask. :)

bicknellr:
I'm going to revert the last commit. I was hoping it was going to make these pass but since it's not helping here I don't see any reason to include it.

usergenic:
if you merge in master you'll get unit and integration tests split out.  maybe easier to see whats going on

bicknellr:
Ok, I'll try that.

usergenic:
Russell, I updated travis.yml on master to reduce sauce session contention and got 4 green travis runs on the first try in a long time.  Merge master again for that goodness.

usergenic:
@bicknellr I went ahead and merged master into this branch to get that effect.  Hopefully will just see green CI in 20 minutes here.

bicknellr:
@usergenic I just saw that PR - awesome! Anxiously awaiting the Travis results: so far 3 for 3.

bicknellr:
Thanks so much for figuring that out!

javatheist:
Pls unpin.

Issue #133, opened 2018-04-13
[bundler] Bundled output contains exports from file URLs


keanulee:
When building https://github.com/PolymerLabs/polymer-redux-hn with `polymer build` and the included polymer.json, build/esm-bundled/src/components/hn-item.js contains this:

```js
import{...,fetchItemIfNeeded,loadFavorites,...,currentItemSelector,...}from"./hn-app.js";
export{fetchItemIfNeeded}from"file:///Users/keanulee/Code/PolymerLabs/polymer-redux-hn/src/actions/items.js";
export{currentItemSelector}from"file:///Users/keanulee/Code/PolymerLabs/polymer-redux-hn/src/reducers/items.js";

...

var hnItem={HnItemElement:HnItemElement,currentItemSelector:currentItemSelector,fetchItemIfNeeded:fetchItemIfNeeded};
export{...,HnItemElement};
```

The source of hn-item.js exports `fetchItemIfNeeded` and `currentItemSelector`, but those two functions are already imported:

```js
import { ..., fetchItemIfNeeded } from '../actions/items.js';
import ..., { currentItemSelector } from '../reducers/items.js';

...

export { currentItemSelector, fetchItemIfNeeded };
```

### Expected

The `export` from file URL statements shouldn't be there as they aren't needed.

usergenic:
Resolved for now.

PR #218, opened 2018-04-24
Rewrite the source value of export declarations so we do not emit file URLs in our output.

usergenic:
There were a few of things that were fixed here.

One small one is, bundler excludes were resulting in rollup errors when imports/exports referenced them.  That was quickly addressed with https://github.com/Polymer/tools/compare/wtf-file-urls?expand=1#diff-87a72e027be8b635fde7aca6c20d749bR52 and that was found and covered now because I used an `excludes` setting in a test to prove the behavior of the focus of the PR, which is...

Second fix is that export sources are rewritten to the bundle URL or otherwise correct relative URL.

Third fix was forgetting to ensure leading dot for relative URL in the import from non-bundled URL case.

### DIFF
--- a/packages/bundler/src/es6-rewriter.ts
+++ b/packages/bundler/src/es6-rewriter.ts
@@ -49,6 +49,7 @@ export class Es6Rewriter {
         external.push(...[...bundle.files, url]);
       }
     }
+    external.push(...this.bundler.excludes);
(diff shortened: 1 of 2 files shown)


usergenic:
Fixes https://github.com/Polymer/tools/issues/133

PR #220, opened 2018-04-25
Scan exported functions

rictic:
They're public and interesting, same as functions on namespaces.

Also use getBestComment in more places, rather than our more ad hoc methods.

This also changes function source ranges a bit, but not in an interesting way IMO.

Part of Polymer/docs#2379

### DIFF
--- a/packages/analyzer/src/javascript/esutil.ts
+++ b/packages/analyzer/src/javascript/esutil.ts
@@ -104,6 +104,18 @@ export function* getSimpleObjectProperties(node: babel.ObjectExpression) {
   }
 }
 
+/** Like getSimpleObjectProperties but deals with paths. */
+export function*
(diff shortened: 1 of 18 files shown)


packages/analyzer/src/javascript/esutil.ts
```diff
@@ -104,6 +104,18 @@ export function* getSimpleObjectProperties(node: babel.ObjectExpression) {
   }
 }
 
+/** Like getSimpleObjectProperties but deals with paths. */
```
aomarks: Can you give me an example to help me understand what a "path" means here?

rictic: A `NodePath` is a richer API wrapped around a babel ast node

PR #234, opened 2018-04-25
Updated travis and appveyor configs to run node 9 instead of latest (aka 10)

usergenic:
This is just the travis and appveyor configs limiting node v to 6 and 9.


rictic:
hm, appveyor isn't building this PR, why is that

usergenic:
I force pushed the same changes again and now its running appveyor.  Must have been a glitch.

PR #235, opened 2018-04-25
Replace all instances of `console.assert()` with `assert()` from 'assert' standard lib.

usergenic:
As of node 10, `console.assert()` doesn't throw an AssertionError anymore.  However `const assert = require('assert'); assert();` does.  So I switched over to `assert()` wherever `console.assert()` was in use.

### DIFF
--- a/packages/analyzer/src/javascript/esutil.ts
+++ b/packages/analyzer/src/javascript/esutil.ts
@@ -12,6 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as assert from 'assert';
 import generate from '@babel/generator';
 import babelTraverse from '@babel/traverse';
 import {NodePath} from '@babel/traverse';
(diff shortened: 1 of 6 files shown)


rictic:
Looks like appveyor is running with node 10 and no longer failing where it was before. Now we've just got a minor whitespace difference in the polymer build integration test goldens. Mind fixing that too in this PR?

rictic:
Hm, weirdly though, node 6 passes on travis without whitespace changes…

usergenic:
Side note: I believe Node 10 has issues (even locally) in polymer-build tests with some dependency of the `temp` library we use to make directories for test scenarios:
```
     Uncaught RangeError [ERR_OUT_OF_RANGE]: The value of "mode" is out of range. Received 16877
      at Object.fs.fchmod (fs.js:1024:11)
      at Object.fchmod (node_modules/graceful-fs/polyfills.js:237:17)
      at mode (node_modules/vinyl-fs/lib/fileOperations.js:132:10)
      at onStat (node_modules/vinyl-fs/lib/fileOperations.js:125:14)
      at node_modules/graceful-fs/polyfills.js:287:18
      at FSReqWrap.oncomplete (fs.js:150:5)
```

I get this locally and on travis for node 10:
screen shot 2018-04-25 at 12 40 24 pm

PR #239, opened 2018-04-25
Remove the custom typing for stable.d.ts that was wrong.

usergenic:
I was wrong and I broke the build.  This fixes the build.

### DIFF
--- a/packages/analyzer/src/model/warning.ts
+++ b/packages/analyzer/src/model/warning.ts
@@ -13,7 +13,7 @@
  */
 
 import * as chalk from 'chalk';
-import stable from 'stable';
+import * as stable from 'stable';
 
 import {Analyzer} from '../core/analyzer';
 import {UrlResolver} from '../index';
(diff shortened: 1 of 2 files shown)


packages/analyzer/src/model/warning.ts
```diff
@@ -13,7 +13,7 @@
  */
 
 import * as chalk from 'chalk';
-import stable from 'stable';
+import * as stable from 'stable';
```
rictic: This file isn't a legal ES6 module, which is why I was using the `import stable = require('stable');` form before.

usergenic:
I'm pretty sure I forgot how to do programming things.

packages/analyzer/src/model/warning.ts
usergenic: Here's the actual fix: https://github.com/Polymer/tools/commit/f8daf698d306140448c1a63439b80a2b2d14f672

usergenic: I am not going to try and monkey with this again.  Sorry!

usergenic: (new ConeOfShame()).apply(usergenic);

PR #243, opened 2018-04-26
Get integration tests happy again

usergenic:
Made this so only one integration server run does the sauces so we have 50% less contention for sauce sessions which means more test-passing goodness.


usergenic:
IT GREENED!

bicknellr:
DO WANT

PR #252, opened 2018-04-26
Allow port to be set when running web-component-tester using --webser…

c-okelly:
Allow port to be set when running web-component-tester using --webserver-port option

Same pull request as https://github.com/Polymer/web-component-tester/pull/687

### DIFF
--- a/packages/web-component-tester/runner/webserver.ts
+++ b/packages/web-component-tester/runner/webserver.ts
@@ -211,6 +211,7 @@ Expected to find a ${mdFilenames.join(' or ')} at: ${pathToLocalWct}/
           componentDir,
           compile: options.compile,
           hostname: options.webserver.hostname,
+          port: options.webserver.port,
           headers: DEFAULT_HEADERS,
           packageName,
           additionalRoutes,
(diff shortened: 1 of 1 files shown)


packages/web-component-tester/runner/webserver.ts
```diff
@@ -261,7 +262,7 @@ Expected to find a ${mdFilenames.join(' or ')} at: ${pathToLocalWct}/
 
     options.webserver._servers = servers.map((s) => {
       const port = s.server.address().port;
-      const hostname = s.options.hostname;
+      const hostname = s.server.address().address;
```
rictic: Why this change?

c-okelly: I thought it would make sense to bring the behaviour of how port and host were set inline with each other. 

I originally set `const port = s.options.port` but this results in a strange default of 0000 when the port isn't set. So I updated hostname to use the same logic as port instead. 

This was probably not really required. Should I remove it ?

rictic: I'd leave it out unless we want to use a library for url construction, because if the address is an ipv6 address then we need to wrap it with square brackets. (i.e. http://[2001:4860:0:2001::68]/ is how you make a url out of the v6 address 2001:4860:0:2001::68])

c-okelly: Changes made.

rictic:
Thanks for the PR!

c-okelly:
Thanks for the review!

Issue #211, opened 2018-04-23
[build/polyserve] AMD transform/requirejs results in multiple requests for same import


usergenic:
So, I *think* this is a client-side behavior of the requirejs wrapper we use that isn't correctly preventing multiple requests for the same file/url.  Repro case is checking out https://github.com/Polymer/polymer/tree/3.x-tests and `polymer serve --module-resolution node --npm` and using Firefox to request http://127.0.0.1:8081/test/unit/utils/resolve-url.js

Have a look at the duplicates in the network tab:
![screen shot 2018-04-23 at 3 23 14 pm](https://user-images.githubusercontent.com/578/39156291-5b577808-470a-11e8-9502-347b4a3b7537.png)

usergenic:
```
diff --git a/test/unit/sub/resolveurl-elements.js b/test/unit/sub/resolveurl-elements.js
index f8219309..cfb18916 100644
--- a/test/unit/sub/resolveurl-elements.js
+++ b/test/unit/sub/resolveurl-elements.js
@@ -12,7 +12,7 @@ import { html } from '../../../lib/utils/html-tag.js';
 import { PolymerElement } from '../../../polymer-element.js';
 import { DomModule } from '../../../lib/elements/dom-module.js';
 import { Polymer } from '../../../lib/legacy/polymer-fn.js';
-import { pathFromUrl } from '../utils/resolve-url.js';
+import { pathFromUrl } from '../../../lib/utils/resolve-url.js';
 const $_documentContainer = document.createElement('div');
 $_documentContainer.setAttribute('style', 'display: none;');
 const baseAssetPath = pathFromUrl(import.meta.url);
```
^ need to do this in order to run the test...

web-padawan:
Probably related to #203, I have also noticed such kind of requests there (Firefox and Edge).

aomarks:
This is highest on my list.

I believe the issue is that RequireJS does not (actually I think _cannot_) resolve relative module dependency URLs the way we assumed. RequireJS always resolves dependencies relative to the document base URL (or an override in a global config). So if `/index.html` loads `"./foo/foo.js"` which then loads `"./bar.js"`, then the absolute URL that is fetched is `/bar.js`, not, as I naively assumed, `/foo/bar.js`. This is because `/foo/foo.js` does not know that it's within the `/foo` path component (incidentally, this is what `import.meta.url` solves).

[...]

Issue #203, opened 2018-04-23
[polyserve] Duplicate CE definition errors in Firefox and Edge when combining paths and bare specifiers in tests


web-padawan:
### Description

Tests for some of the Vaadin components are failing, because of external dependency usage. They pass in Chrome and Safari, but produce the errors like this in Firefox and Edge:

```
"Error: A custom element with name 'vaadin-text-field' has already been defined."
````
Most likely caused by not deduplicating requests originated from modulizer output:
- `<script type="module" src="../../../@vaadin/vaadin-text-field/vaadin-text-field.js"></script>`
- `import '@vaadin/vaadin-text-field/vaadin-text-field.js';`

### Versions & Environment
- Polymer CLI: 1.7.0-pre.13
- node: 8.11.1
- Operating System: Mac OS

### Steps to Reproduce

1. Clone the repo:
```sh
git clone --depth 1 -b p3-preview-bare git://github.com/vaadin/vaadin-list-box.git && cd vaadin-list-box
```

2. Install the modulizer:
```sh
git clone git://github.com/Polymer/polymer-modulizer.git && cd polymer-modulizer && npm link
```

3. Install the tools:
```sh
npm i -g bower yarn magi-cli polymer-cli@next
```

4. Install bower dependencies:
```sh
bower install
```

5. Reset the last commit (containing the workaround):
```
git reset --hard HEAD^1
```

6. Run the conversion:
```sh
magi p3-convert --out . --import-style=name
```

7. Install dependencies:
```sh
yarn installl --flat
```

8. Serve the component:
```sh
polymer serve --npm --module-resolution=node
``` 

9. Open the Firefox and go to http://127.0.0.1:8081/components/@vaadin/vaadin-list-box/test/

### Workaround

We managed to solve this by putting the external HTML imports inside of the different file, then importing that one in test suite, in order to get same paths for HTML and JS. See it here:
- vaadin/vaadin-list-box@0ecfee4 - [Travis green](https://travis-ci.org/vaadin/vaadin-list-box/jobs/370042023)
- vaadin/vaadin-form-layout@6cedbc0 - [Travis green](https://travis-ci.org/vaadin/vaadin-form-layout/jobs/370037928)

### Expected Results (with the workaround):

Tests passing, see the builds above and the screenshot:

![screen shot 2018-04-23 at 14 10 46](https://user-images.githubusercontent.com/10589913/39123872-f6ba90b8-4701-11e8-8400-fbf8e71814cf.png)

[...]

PR #254, opened 2018-04-26
New AMD-style module loader.

aomarks:
This is a replacement for our use of RequireJS to use as the loader for our ES modules transform. It is AMD-ish, but not 100% compliant with the AMD spec or RequireJS.

Differences from RequireJS:
- 0.9 KB minified + compressed vs 6.6 KB for RequireJS.

- Correctly resolves relative paths above the base URL (see https://github.com/Polymer/tools/issues/211 and possibly https://github.com/Polymer/tools/issues/203).

- Is entirely path-based. Does not allow registering a module by name. The only API is `define(deps, factory)`.

- `define()` module factories are always executed. RequireJS only starts executing a module dependency graph when triggered with a `require()`. The Babel AMD transform only emits `define()` calls, so this eliminates the need for us to inject a `require()` call during build.

- Identifies top-level scripts and ensures they execute in order, like `type=module` scripts do. This would eliminate some complexity in our build pipeline, which does the same thing using code generation, and has to do some ugly tricks to preserve script-order information through the `HtmlSplitter` process.

- Modules can request a `meta` dependency, which is resolved to `{url: <moduleUrl>}`. This should be able to simplify our build-time `import.meta` transform, and has the advantage of not being tied to a particular static path (i.e. you could move the module around your tree post-build and still get a correct URL).

- Modules always provide an `exports` object to their dependents, even if nothing was exported from the module, which is more similar to native ES modules. This would fix https://github.com/Polymer/polymer-cli/issues/995 in the general case.


packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+    this.notify[i]();
+  }
+};
+
+Module.resolverAnchor = document.createElement('a');
+
+Module.prototype.resolveUrl = function(path) {
+  if (path.indexOf('://') !== -1) {
+    // Already a fully qualified URL.
+    // TODO(aomarks) Is this a good enough check?
```
bicknellr: If you have `URL`, then I think you can try-catch `new URL(path)`, which throws if it's not given a full URL, IIRC.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+    return path;
+  }
+  // Just let the browser do path resolution/normalization.
+  Module.resolverAnchor.href = this.basePath + path;
```
bicknellr: Also, if you have `URL`, `(new URL(path, this.basePath)).href` is safer since it handles joining them properly:

```
> "http://example.com/page" + "some/path.js"
< "http://example.com/pagesome/path.js"
```

... when really it should be:

```
> (new URL("some/path.js", "http://example.com/page")).href
< "http://example.com/some/path.js"
```

(`this.basePath` needs to be a full URL though.)

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+
+function getBasePath(url) {
+  // TODO(aomarks) Maybe a better way to do this.
+  var lastSlash = url.lastIndexOf('/');
+  if (lastSlash === url.indexOf('://') + 2) {
+    // http://example.com -> http://example.com/
+    return url + '/';
+  }
+  // http://example.com/foo/bar.html -> http://example.com/foo/
+  return url.substring(0, lastSlash + 1);
```
bicknellr: Also, if you have `URL`: `const getBasePath = url => new URL("./", url).href;`

packages/amd-loader/README.md
```diff
@@ -0,0 +1,5 @@
+[![NPM version](http://img.shields.io/npm/v/@polymer/amd-loader.svg)](https://www.npmjs.com/package/@polymer/amd-loader)
+
+# @polymer/amd-loader
+
+A JavaScript library that loads AMD-style modules.
```
rictic: Let's add some text saying that it loads path style modules and not name style modules, and that it's intended to mimic the semantics of es6 modules

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
+'use strict';
+
+(function() {
+
+function Registry() {
+  this.modules = {};
```
rictic: Use `Object.create(null)` for a map-style object, otherwise stuff like `this.modules['valueOf']` will be truthy.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+  this.topLevelIdx = 0;
+  this.previousTopLevelUrl = undefined;
+  this.pendingDefine = undefined;
```
rictic: Document these fields please. Maybe use `// @ts-check` and jsdoc to get basic type checking, or use full typescript and tsc can also handle transpiling to ES5

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+  return module;
+};
+
+var registry = new Registry();
+
+function getBasePath(url) {
+  // TODO(aomarks) Maybe a better way to do this.
+  var lastSlash = url.lastIndexOf('/');
+  if (lastSlash === url.indexOf('://') + 2) {
```
rictic: What about urls like `//example.com`?

packages/amd-loader/amd-loader.js
rictic: Worth noting that IE 11 does not have URL, but the webcomponents polyfill includes it.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+(function() {
+
+function Registry() {
+  this.modules = {};
+  this.topLevelIdx = 0;
+  this.previousTopLevelUrl = undefined;
+  this.pendingDefine = undefined;
+}
+
+Registry.prototype.get = function(url) {
```
rictic: Document these methods as well please. For most of them, just the types of the params and return is enough, but for ones with complicated invariants they're maintaining those would be valuable as well.

justinfagnani:
A few high-level things:

- @azakus apparently added the URL polyfill to the platform layer of the webcomponents polyfills, so it should be ok to use it.
- If we consider this to be more of a polyfill for modules, I wonder if it should be loaded as part of the webcomponents polyfills? <- @sorvell @kevinpschaaf
- It'd be nice to write this in TypeScript to get the type-checking benefits we're used to. TODO'able since we want to not spend too much time on this now.

packages/amd-loader/README.md
justinfagnani: Yes, let's call out that it's a subset and mention the differences you have in the PR description.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
+'use strict';
+
+(function() {
+
+function Registry() {
```
justinfagnani: Does `Registry` need to be a "class" at all? If it's a singleton you can just put the state and functions in the module closure.

packages/amd-loader/amd-loader.js
justinfagnani: +1 on TypeScript

packages/amd-loader/amd-loader.js
justinfagnani: Even though what's inside the module closure is top-level, I'd still indent it. Initially I thought some things below were outside of the IIFE for some reason.

aomarks:
> @azakus apparently added the URL polyfill to the platform layer of the webcomponents polyfills, so it should be ok to use it.

Seems like with some unit tests we can get by with a built-in function that just has to find a base path from a URL, no? The URL polyfill is bigger than this entire library, and in theory this loader could be useful to people who don't otherwise need any webcomponents polyfills.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+
+  var thisResolve = this.resolve.bind(this);
+  function maybeResolve() {
+    if (numUnresolvedDeps === 0) {
+      factory.apply(null, factoryArgs);
+      thisResolve();
+    }
+  }
+
+  for (var i = 0; i < (deps || []).length; i++) {
```
justinfagnani: `(deps || []).length` will be evaluated every iteration

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+    var depSpec = deps[i];
+
+    if (depSpec === 'exports') {
+      factoryArgs[i] = this.exports;
+
+    } else if (depSpec === 'require') {
+      factoryArgs[i] = this.makeDynamicImporter();
+
+    } else if (depSpec === 'meta') {
```
justinfagnani: AMD includes a "module" built-in dependency, but the spec doesn't seem to specify it that I can see. In IMD it has the `exports` and `id` properties. Maybe we dug that out of RequireJS source. Seems like in this loader `module.id` might === `import.meta.url`.

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+    } else {
+      // The script did not make a call to define(), otherwise the global
+      // callback would have been set. That's fine, we can resolve it
+      // immediately, because it doesn't have any dependencies.
+      module.resolve();
+    }
+  };
+
+  script.onerror = function() {
+    throw new Error('error loading module', url);
```
justinfagnani: does this need to propagate to dependents?

packages/amd-loader/amd-loader.js
```diff
@@ -0,0 +1,196 @@
...
+    module = this.modules[url] = new Module(url);
+  }
+  return module;
+};
+
+var registry = new Registry();
+
+function getBasePath(url) {
+  // TODO(aomarks) Maybe a better way to do this.
+  var lastSlash = url.lastIndexOf('/');
```
justinfagnani: This might have issues with slashes in query parameters or fragments. They are possible in module specifiers.

packages/amd-loader/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@polymer/amd-loader",
```
justinfagnani: We might want a name that indicates that this isn't full AMD: `amd-path-loader`?

packages/amd-loader/amd-loader.js
aomarks: Added handling for those cases. Will cover in unit tests.

packages/amd-loader/amd-loader.js
aomarks: Looks like the AMD spec defers to the CommonJS spec for `module`, which says:

> 1) The "module" object must have a "id" property that is the top-level "id" of the module. The "id" property must be such that require(module.id) will return the exports object from which the module.id originated. (That is to say module.id can be passed to another module, and requiring that must return the original module). When feasible this property should be read-only, don't delete.
> 
> 2) The "module" object may have a "uri" String that is the fully-qualified URI to the resource from which the module was created. The "uri" property must not exist in a sandbox.
> 
> http://wiki.commonjs.org/wiki/Modules/1.1.1#Module_Context

I'm open to using `module.id` or `module.uri` (or both). Although, I do also like the idea of providing it via `meta`, because we are trying to match the semantics of `import.meta`. I don't know if `import.meta.scriptElement` is going to happen, but if it did, then we would have a natural place to put it. I think this discussion depends on what our position is on interopability with other AMD modules/loaders.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+   * which is why this is not simply part of construction.
+   */
+  define(deps: string[], factory?: (...args: {}[]) => void) {
+    const mod = this;
+    this.require(deps, function(...args: {}[]) {
+      if (factory !== undefined) {
+        factory.apply(null, args);
+      }
+      mod.resolved = true;
+      for (const callback of mod.notify) {
```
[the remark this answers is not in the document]
aomarks: Added `downlevelIteration: false`.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Load this module by creating a <script> tag in the document <head>, unless
+   * we have already started (or didn't need to, as in the case of top-level
+   * scripts).
+   */
+  private loadIfNeeded() {
+    if (this.needsLoad === false) {
+      return;
+    }
+    this.needsLoad = true;
```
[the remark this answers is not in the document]
aomarks: Fixed.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+
+    const script = document.createElement('script');
+    script.src = this.url;
+
+    const mod = this;
+    script.onload = function() {
+      if (pendingDefine !== undefined) {
+        pendingDefine(mod);
```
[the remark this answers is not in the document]
aomarks: Discussed offline. The timing doesn't work out -- the module object has already been created by the parent module at this time (whichever one first declared a dependency on this module), and here we are actually waiting to find out our URL so that we can associate that existing module object with the dependencies and factory that we now have.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+ * A global map from a fully qualified module URLs to module objects.
+ */
+const registry: {[url: string]: Module} = Object.create(null);
+
+/**
+ * Return a module object from the registry for the given URL, creating one if
+ * it doesn't exist yet.
+ */
+function getModule(url: string): Module {
+  let mod = registry[url];
```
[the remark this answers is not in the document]
aomarks: We know that the keys are all fully qualified URLs (now with a branded type to be really clear), so I don't think this is a problem.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+class Module {
+  private url: normalizedUrl;
+  private urlBase: string;
+  private exports: {[id: string]: {}} = {};
+
+  /** All dependencies are resolved and the factory has run. */
+  private resolved = false;
+
+  /** Callbacks from dependents waiting for this module to resolve. */
+  private notify: Array<() => void> = [];
```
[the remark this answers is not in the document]
aomarks: Discussed offline. The current factoring exploits the fact that the same implementation of `require` can be used both for module initialization ("define"), and as the function we pass to modules requesting a dynamic "require". However, in the latter case, we don't have any dependents to notify, so that happens only in "define" with an additional closure.

packages/amd-loader/amd-loader.js
aomarks: Also got rid of the `Module` class, from offline discussion. Size is now down to 0.7KB :)

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+  // script executing this define() call could have been loaded:
+
+  // Case #1: We are a dependency of another module. A <script> was injected
+  // to load us, but we don't yet know the URL that was used. Because
+  // document.currentScript is not supported by IE, we communicate the URL via
+  // a global callback. When finished executing, the "onload" event will be
+  // fired by this <script>, which will be handled by the loading script,
+  // which will invoke the callback with our module object.
+  let defined = false;
+  pendingDefine = (mod) => {
```
usergenic: Talked about this offline, but I was concerned here that this was overwriting a global singleton and that doing so invites, potentially, a tricky-to-diagnose error condition if there is a possibility of two consecutive `define()` invocations before the triggering of the `pendingDefine()` inside `script.onload`.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  function checkDeps() {
+    if (numUnresolvedDeps === 0) {
+      if (callback !== undefined) {
+        callback.apply(null, args);
+      }
+    }
+  }
+
+  for (let i = 0; i < deps.length; i++) {
```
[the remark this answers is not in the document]
aomarks: Done.

aomarks:
I just pushed some WCT tests, and some bug fixes that they revealed. PTAL!

The major remaining piece here is error handling, e.g. it doesn't yet support invoking the error callback for dynamic require calls when a dependency cannot be resolved.

packages/amd-loader/.gitignore
```diff
@@ -0,0 +1,2 @@
+/node_modules/
+/lib/
```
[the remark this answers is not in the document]
justinfagnani: Yep. Empty .npmignore will do

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,237 @@
+type modCallback = (...args: {}[]) => void;
```
justinfagnani: capitalize

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,237 @@
...
+      exports: Object.create(null),
+      resolved: false,
+      needsLoad: true,
+      notify: [],
+    };
+  }
+  return mod;
+}
+
+type normalizedUrl = string&{_normalized: never};
```
justinfagnani: capitalize and move to top of file

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,237 @@
+type modCallback = (...args: {}[]) => void;
+
+interface Window {
+  define: (deps: string[], factory: modCallback) => void;
+}
+
+(function() {
```
justinfagnani: you shouldn't need an IIFE in the source - either the module setting in tsconfig will create one, or just use a block, which should be compiled to an IIFE.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * Examples:
+ *
+ *  - http://example.com/ => http://example.com/
+ *  - http://example.com/foo.js => http://example.com/
+ *  - http://example.com/foo/ => http://example.com/foo/
+ *  - http://example.com/foo/?qu/ery#fr/ag => http://example.com/foo/
+ */
+function getUrlBase(url: normalizedUrl): normalizedUrl {
+  url = url.split('?')[0] as normalizedUrl;
```
justinfagnani: can you use the anchor for this?

packages/amd-loader/README.md
aomarks: Just pushed a much more complete README, PTAL.

packages/amd-loader/src/amd-loader.ts
justinfagnani: The default module setting is CommonJS, which might not be obvious in the output here because you have no imports or exports. `"module": "none"` and using a block might do it.

justinfagnani:
Do you have an idea for a new name that better hints at the specific functionality?

packages/amd-loader/src/amd-loader.ts
aomarks: Nope, module: none and target:es5 with a block will just emit a block with `var`s:

tsconfig.json
```json
{
  "compilerOptions": {
    "target": "es5",
    "module": "none"
  }
}
```

test.ts
```ts
{
  let x = 0;
}
```

test.js
```js
{
  var x = 0;
}
```

aomarks:
> Do you have an idea for a new name that better hints at the specific functionality?

Does seem like we need something, but nothing so far sounds great:

amd-path-loader
tiny-amd-loader
require-less
amd-loader-lite
es-module-amd-loader

Any other ideas?

aomarks:
> Yep. Empty .npmignore will do

Done. Needed some more stuff in .npmignore anyway.

packages/amd-loader/src/amd-loader.ts
justinfagnani: Wow, that just seems broken... unless the rationale is without a module setting they can't know that this file _isn't_ the whole program?

Ok, then. Thanks for humoring me! 🤷‍♂️

justinfagnani:
amd-path-loader isn't bad, IMO. js-module-loader?

packages/amd-loader/amd-loader.js
aomarks: Yes, and this now happens.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+    this.url = normalizeUrl(url);
+    this.urlBase = getBasePath(this.url);
+  }
+
+  /**
+   * Initialize this module with its dependencies and factory function. Note
+   * that Module objects are created and registered before they are loaded,
+   * which is why this is not simply part of construction.
+   */
+  define(deps: string[], factory?: (...args: {}[]) => void) {
```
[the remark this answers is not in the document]
aomarks: We think not.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+    for (let i = 0; i < deps.length; i++) {
+      const depSpec = deps[i];
+
+      if (depSpec === 'exports') {
+        args[i] = this.exports;
+
+      } else if (depSpec === 'require') {
+        args[i] = this.require.bind(this);
+
+      } else if (depSpec === 'meta') {
```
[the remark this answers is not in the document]
aomarks: Going with meta for now.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+    } else {
+      // The script did not make a call to define(), otherwise the global
+      // callback would have been set. That's fine, we can resolve immediately
+      // because we don't have any dependencies, by definition.
+      define(mod, []);
+    }
+  };
+
+  script.onerror = () => {
+    throw new Error('error loading module ' + mod.url);
```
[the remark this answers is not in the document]
aomarks: No, but now it does because there is propagation.

packages/amd-loader/src/amd-loader.ts
```diff
@@ -0,0 +1,242 @@
...
+        depMod.notify.push(onDepResolved);
+        loadIfNeeded(mod);
+      }
+    }
+  }
+
+  checkDeps();
+}
+
+let pendingDefine: ((mod: Module) => void)|undefined = undefined;
```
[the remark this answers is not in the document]
aomarks: I rearranged the window.define function to the top so that it is closer to this variable.

packages/amd-loader/src/amd-loader.ts
aomarks: Done.

packages/amd-loader/src/amd-loader.ts
aomarks: Done.

packages/amd-loader/src/amd-loader.ts
aomarks: `.pathname` could be useful, but then we need to reconstruct scheme, host, etc., so we think this is simpler for now.

Issue #253, opened 2018-04-26
[build] Miscompile of 3.0 version of shop - caseMap$1 is not defined


rictic:
Title is vague as I'm still investigating, but loading the es5-bundled version of shop 3.0 built with polymer-cli at master throws an error because we reference an undefined variable `caseMap$1` from the transformed output of `@polymer/polymer/lib/mixins/property-effects.js`

I think the issue is with more than just that symbol though, because if I work around that failure there are other references in that file that are similarly miscompiled. Still investigating.

rictic:
It looks like the issue is in a loop rewriter that's mangling the while loop near the top of `property-effects.js`. Minimal repro, this code:

```js
import './empty.js';

let foo;
while (0);

console.log(foo);
```

Is rewritten to:

```js
console.log(foo);
```

The `foo` variable declaration is being deleted, so this is an error.

That's what's happening in the original error, the caseMap and nativeProperties local variable declarations are deleted, causing the error.

rictic:
Disabling the `simplify` babel-minify plugin helps. Filing a bug against that repo.

rictic:
Filed as https://github.com/babel/minify/issues/824

PR #255, opened 2018-04-26
Disable the `simplify` babel plugin when minifying js

rictic:
See https://github.com/babel/minify/issues/824

Fixes https://github.com/Polymer/tools/issues/253


aomarks:
Regression test would be nice if there's an obvious way to do it, similar to https://github.com/Polymer/tools/blob/master/packages/build/src/test/js-transform_test.ts#L341

rictic:
Added regression test and verified that it failed before this change.

PR #259, opened 2018-04-27
Prepare dev releases.

aomarks:
A set of dev-releases so that some folks can test the new AMD loader.

(Note this is a merge into the amd-build branch).


rictic:
I think dev releases should:
  - be labeled, not numbered
  - not be checked into master

And maybe they don't need to go through code review?

rictic:
I'd like to do nightly automated releases, but we don't have time for that right now I don't think.

What about recommending users build from master?

PR #270, opened 2018-04-28
Add JS language levels as options to builds.

### LINKED ISSUE #173: Edge supports ES6, but not object spread operator
Currently we have the `js.compile` option which is either `true` (es5 output) or `false` (es6 output). While Edge supports es6, it does not support newer syntax such as object spread. We either need to add more granular JS build settings or update browser capabilities for Edge.

justinfagnani:
By not running the es2015 related transforms on the 4 current major browsers, we eliminate the need for regenerator runtime on them, but still support object spread, async iterators, etc.

Fixes #173 and Polymer/polymer-cli#1000

We'll need another change to inject regenerator runtime for IE11.


packages/browser-capabilities/src/browser-capabilities.ts
```diff
@@ -44,12 +53,19 @@ const browserPredicates: {
...
     es2015: since(36),
+    es2016: since(45),
+    es2017: since(45),
+    es2018: since(51),
     push: since(28),
     serviceworker: since(32),
     modules: since(48),
   },
   'Vivaldi': {
+    //
```
aomarks: remove comment

packages/project-config/src/test/builds_test.ts
```diff
@@ -28,20 +27,25 @@ suite('builds', () => {
       assert.equal(isValidPreset('es6'), false);
       assert.equal(isValidPreset('js-compile'), false);
       assert.equal(isValidPreset(''), false);
-      assert.equal(isValidPreset(null), false);
-      assert.equal(isValidPreset(undefined), false);
-      assert.equal(isValidPreset(0), false);
-      assert.equal(isValidPreset(1), false);
+      assert.equal(isValidPreset(null as any), false);
```
rictic: this should fail to lint. need to add

`// tslint:disable-next-line: no-any Explicitly testing invalid code.`

packages/project-config/src/test/index_test.ts
```diff
@@ -551,23 +539,23 @@ suite('Project Config', () => {
 
   suite('loadOptionsFromFile()', () => {
     test('throws an exception for polymer.json with invalid syntax', () => {
-      const filepath = path.join(__dirname, 'polymer-invalid-syntax.json');
+      const filepath = path.resolve('test/polymer-invalid-syntax.json');
```
rictic: Can we assume the cwd in tests? I'd been going off of the assumption that file-relative paths with `__dirname` are safer.

packages/project-config/src/builds.ts
```diff
@@ -125,16 +129,18 @@ export interface ProjectBuildOptions {
...
       exclude?: string[],
     },
 
     /** Use babel to compile all ES6 JS down to ES5 for older browsers. */
-    compile?: boolean|{
-      /** JavaScript files listed here will not be compiled. */
-      exclude?: string[],
-    },
+    compile?:
+               boolean|JsCompileTarget|{
```
rictic: Is this backwards compatible? Will there be problems when a user runs an old version of polymer cli with a string for `compile`?

packages/polyserve/src/compile-middleware.ts
```diff
@@ -85,7 +85,7 @@ export function babelCompile(
     transform(request: Request, response: Response, body: string): string {
       const capabilities = browserCapabilities(request.get('user-agent'));
       const options = {
-        transformES2015: compile === 'always' || !capabilities.has('es2015'),
+        compile: compile === 'always' || !capabilities.has('es2015'),
```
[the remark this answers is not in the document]
aomarks: `browserCapabilities(userAgent).latestEsVersion()` or something might be nice

packages/polyserve/src/compile-middleware.ts
```diff
@@ -84,8 +87,24 @@ export function babelCompile(
...
     transform(request: Request, response: Response, body: string): string {
       const capabilities = browserCapabilities(request.get('user-agent'));
+      let compileTarget: JsCompileTarget = undefined;
+      if (compile === 'always') {
+        compileTarget = 'es5';
+      } else if (compile === 'auto') {
+        if (capabilities.has('es2018')) {
+          compileTarget = undefined;
+        } else if (capabilities.has('es2017')) {
+          compileTarget = 'es2018';
```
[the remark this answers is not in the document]
justinfagnani: Fixed

packages/project-config/src/test/index_test.ts
justinfagnani: Seems we can. I've run this test from lerna from the root and local to the package with npm.

packages/browser-capabilities/src/browser-capabilities.ts
justinfagnani: done

justinfagnani:
@aomarks @rictic PTAL

PR #275, opened 2018-04-30
Add full integration test for 3.0 version of shop.

rictic:
Fixes https://github.com/Polymer/tools/issues/137


packages/cli/src/test/integration/integration_test.ts
```diff
@@ -117,14 +133,70 @@ suite('integration tests', function() {
...
-      //   {cwd: dir})
-      // await runCommand(binPath, ['test'], {cwd: dir})
-      await runCommand(binPath, ['build'], {cwd: dir});
+      await Promise.all([
+        // Does not lint clean at the moment.
+        // TODO: https://github.com/Polymer/tools/issues/274
+        // runCommand(binPath, ['lint', '--rules=polymer-3'], {cwd: dir}),
+        runCommand(binPath, ['build'], {cwd: dir}),
+      ]);
+      async function assertTrueInPage(
```
aomarks: Can you add some newlines between function definitions? kinda dense

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -117,14 +133,70 @@ suite('integration tests', function() {
...
+          page: puppeteer.Page, expression: string) {
+        assert(
+            await page.evaluate(expression),
+            `Expected \`${expression}\` to evalutate to true in the browser.`);
```
aomarks: evaluate

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -117,14 +133,70 @@ suite('integration tests', function() {
...
+        await page.waitFor(function(this: Window) {
+          return this.customElements.get('shop-cart') !== undefined;
+        });
+      }
+      const config =
+          ProjectConfig.loadConfigFromFile(path.join(dir, 'polymer.json'));
+      if (config == null) {
+        throw new Error('Failed to load shop\'s polymer.json');
+      }
+      await Promise.all(config.builds.map(
```
aomarks: Looks like it's not going to be obvious which build was running when any assertion failed, is that right? Would be nice to define a different test for each build, although I guess then it won't run in parallel. Update all the assert calls to mention the build or something?

rictic: Yeah the most important thing is really just avoiding repeating the build action, but it's a bit faster to run the browser tests in parallel

packages/analyzer/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Dropped support for node v6. This is a soft break, as we aren't
```
justinfagnani: Reference our node support policy.

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -31,6 +35,18 @@ suite('integration tests', function() {
 
   // Extend timeout limit to 90 seconds for slower systems
   this.timeout(120000);
+  let browserPromise: Promise<puppeteer.Browser>;
+  const disposables: Array<() => void | Promise<void>> = [];
+  before(() => {
```
justinfagnani: Don't use before with suite/test, it has weird behavior. Use setup/teardown

PR #285, opened 2018-05-01
Refactor esm-amd-loader test into its own package.

aomarks:
This breaks the circular dependency between esm-amd-loader and web-component-tester, which causes issues with lerna since it can't know which order to build in.


packages/esm-amd-loader/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Minimal AMD-style loader for replicating ES module behavior.",
```
rictic: Let's add /test/ to esm-amd-loader/.npmignore to avoid any issues with having multiple nested package.json files in our published sources for this repo.

aomarks: It's already there.

Issue #294, opened 2018-05-02
Add import.meta transform in Bundler


justinfagnani:
The Bundlers assumes modules in and modules out, but by moving the location of files it changes the answer of `import.meta`. The Bundler should have it's own `import.meta` transform that can assume `import.meta` support in the output, but just fixes up the `.url` property so that it returns the original file URL.

So

`import.meta`

should become:

`{...import.meta, url: new URL(relativePathFromBundleToOriginal, import.meta.url).toString()}`

web-padawan:
Do I understand correctly that with this fix we will not be able to check `import.meta.url` to detect whether the app is bundled or not as I was assuming, e. g. like this:

```js
let isUnbundled = this.importPath.endsWith(`${this.constructor.is}.js`);
```
Apparently my assumption was wrong, so I'll close #204 in favor of this issue.

There is a regression in comparison with HTML imports, though: we can do recursive HTML imports check and detect whether the specific import has been loaded or not, but AFAIK that is not possible with the current native ES modules implementation.

justinfagnani:
> Do I understand correctly that with this fix we will not be able to check import.meta.url to detect whether the app is bundled or not

Correct. `import.meta.url` is used in Polymer elements to resolve URLs to assets correctly. It needs to return the original URL of the module.

We're considering adding something like `import.meta._bundleUrl` to point to the actual bundled module's URL.

> There is a regression in comparison with HTML imports, though: we can do recursive HTML imports check and detect whether the specific import has been loaded or not, but AFAIK that is not possible with the current native ES modules implementation.

I don't quite understand this. What check could you do with HTML imports?

web-padawan:
@justinfagnani there is the code we're currently using: https://github.com/vaadin/vaadin-development-mode-detector/blob/master/vaadin-development-mode-detector.html#L17

The idea is that, if the app is bundled, then Polymer should be included into some bundle (usually the app shell, when using CLI) and the `polymer-element.html` should not be imported.

justinfagnani:
That seems very unnecessary. Bundler will take care of rewriting imports to the correct bundle, so you don't need to remove the import of `polymer-element.html`.

usergenic:
https://github.com/Polymer/tools/pull/300 <- should address

web-padawan:
@justinfagnani the code which I referenced above is not meant to remove the import, but to detect whether the app is bundled or not, based on it's presence.

usergenic:
[...]

PR #300, opened 2018-05-02
Preserve correctness of `import.meta.url` in bundled modules.

usergenic:
Note that in the case where an imported module's URL is identical to the bundle URL, no modification to the module code is needed before bundling it, but for all other cases where `import.meta` occurs somewhere in the document contents, we construct a module-scoped stand-in variable and swap other references to import.meta to use the stand-in, which is declared as an object-literal like `{...import.meta, url: new URL(relativeUrlToModule, import.meta.url)}`.


usergenic:
Fixes https://github.com/Polymer/tools/issues/294

packages/bundler/src/es6-rewriter.ts
```diff
@@ -106,20 +120,37 @@ export class Es6Rewriter {
...
+                  !document.parsedDocument.contents.includes('import.meta')) {
+                return document.parsedDocument.contents;
+              }
+
+              // We need to rewrite instances of `import.meta` in the document
+              // to preserve its location because `import.meta` is used and the
+              // URL has changed as a result of bundling.
+              const relativeUrl =
+                  ensureLeadingDot(this.bundler.analyzer.urlResolver.relative(
+                      this.bundle.url, id as ResolvedUrl));
```
rictic: Is the cast of `id` needed? It looks like it's a ResolvedUrl already

packages/bundler/src/es6-rewriter.ts
```diff
@@ -106,20 +120,37 @@ export class Es6Rewriter {
...
+              const newAst = this._rewriteImportMetaToBundleMeta(
+                  document.parsedDocument.ast as babel.File, relativeUrl);
```
rictic: We could put this constraint into the type system if we added a type checking function that checks that `document` is a `Document<ParsedJavascriptDocument>`

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +463,52 @@ export class Es6Rewriter {
...
+        babel.variableDeclaration(  // leave this in for clang-format
+            'const',
+            [babel.variableDeclarator(
+                babel.identifier(bundledImportMetaName),
+                babel.objectExpression([
+                  babel.spreadProperty(babel.memberExpression(
+                      babel.identifier('import'), babel.identifier('meta'))),
+                  babel.objectProperty(
+                      babel.identifier('url'),
+                      babel.newExpression(
```
rictic: We want to get the `.href` property of this new url, as `import.meta.url` is a string

packages/bundler/src/test/import-meta-url_test.ts
```diff
@@ -0,0 +1,108 @@
...
+      import {aUrl, myUrl as bUrl} from './subfolder/b.js';
+      const myMeta = import.meta;
+      export {aUrl, bUrl, myMeta};
+    `,
+  });
+
+  const aUrl = analyzer.resolveUrl('a.js')!;
+  const bUrl = analyzer.resolveUrl('subfolder/b.js')!;
+  const cUrl = analyzer.resolveUrl('c.js')!;
+  const dUrl = analyzer.resolveUrl('d.js')!;
```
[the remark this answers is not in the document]
usergenic: No, thanks!  I was adding another test case and then gave up; forgot that was there.  Removed.

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +463,52 @@ export class Es6Rewriter {
         importSpecifier,
         {type: 'ImportSpecifier', imported: babel.identifier(exportName)});
   }
+
+  private _rewriteImportMetaToBundleMeta(
+      moduleFile: babel.File,
+      relativeUrl: FileRelativeUrl): babel.File {
+    // Generate a stand-in for any local references to import.meta...
+    // const __bundledImportMeta = {...__importMeta, url: __bundledImportURL};
```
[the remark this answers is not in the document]
usergenic: I previously had a local variable stand-in just for import.meta that is now gone. Old comment. Fixed to import.meta! Thanks!

packages/bundler/src/es6-rewriter.ts
usergenic: You're right. Fixed.

packages/bundler/src/es6-rewriter.ts
usergenic: You're right and in fact this code probably should have more guarantees/verification of the type of document it is loading anyways. I'll add a TODO here for this.

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +463,52 @@ export class Es6Rewriter {
...
+    const bundledImportMetaName = '__bundledImportMeta';
+    const bundledImportMetaDeclaration =
```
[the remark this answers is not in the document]
usergenic: I love this idea.  I even made a personal TODO note earlier today "Look up to see if there's a template string builder that would make this code cleaner."

I would like to consider bringing that approach to all of the ast production in bundler, but I'm going to hold off on a separate PR to bring that machinery in here.  Adding TODO.  It's a great idea and would make this code a lot better.

I should note that as of right now though, Bundler does not have any babel parsing in it and relies purely on Analyzer. Use of the template literal stuff would essentially require ensuring I either find a nice way to clone the Analyzer's babel parsing stack or rebuild it locally.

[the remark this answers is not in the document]
aomarks: SGTM

packages/bundler/src/es6-rewriter.ts
usergenic: DERP!  Really good catch.  Fixed.

usergenic:
@rictic & @aomarks comments addressed

usergenic:
Awww nuts.  Something probably in the package-lock has updated causing something to break during gulp-typescript... https://ci.appveyor.com/project/Polymer/tools/build/1.0.858#L127

usergenic:
Bunch of things like `Property '_transform' in type 'GulpTransform' is not assignable to the same property in base type 'Transform'.`

usergenic:
Ah this failure is unrelated to branch.  Created a new PR to start to fix the TS errors related to updating package locks at https://github.com/Polymer/tools/pull/309

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +465,60 @@ export class Es6Rewriter {
...
         {type: 'ImportSpecifier', imported: babel.identifier(exportName)});
   }
+
+  private _rewriteImportMetaToBundleMeta(
+      moduleFile: babel.File,
+      relativeUrl: FileRelativeUrl): babel.File {
+    // Generate a stand-in for any local references to import.meta...
+    // const __bundledImportMeta = {...import.meta, url: __bundledImportURL};
+    // TODO(usergenic): Consider migrating this AST production mishmash into the
+    // `ast` tagged template literal available like this:
```
justinfagnani: +1. Much easier to read and write.

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +465,60 @@ export class Es6Rewriter {
...
+    // https://github.com/Polymer/tools/blob/master/packages/build/src/babel-plugin-dynamic-import-amd.ts#L64
+    const bundledImportMetaName = '__bundledImportMeta';
```
justinfagnani: We should generate this name to be conflict free with identifiers at `import.meta` use sites.

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +465,60 @@ export class Es6Rewriter {
...
+              babel.identifier(bundledImportMetaName), babel.objectExpression([
+                babel.spreadProperty(babel.memberExpression(
+                    babel.identifier('import'), babel.identifier('meta'))),
+                babel.objectProperty(
+                    babel.identifier('url'),
+                    babel.memberExpression(
+                        babel.newExpression(
+                            babel.identifier('URL'),
+                            [
+                              //
```
justinfagnani: remove? Or if it's for formatting, move to previous line?

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +465,60 @@ export class Es6Rewriter {
...
+                              babel.memberExpression(
+                                  babel.memberExpression(
+                                      babel.identifier('import'),
+                                      babel.identifier('meta')),
+                                  babel.identifier('url'))
+                            ]),
+                        babel.identifier('href')))
+              ]))
+        ]);
+    const newModuleFile = clone(moduleFile);
```
justinfagnani: why do you have to clone it?

packages/bundler/src/es6-rewriter.ts
```diff
@@ -432,4 +465,60 @@ export class Es6Rewriter {
...
+        enter(path: NodePath<babel.MetaProperty>) {
+          const metaProperty = path.node;
+          if (metaProperty.meta.name !== 'import' &&
+              metaProperty.property.name !== 'meta') {
+            // We're specifically looking for instances of `import.meta` so
+            // ignore any other meta properties.
+            return;
+          }
+          const bundledImportMeta = babel.identifier(bundledImportMetaName);
+          rewriteObject(metaProperty, bundledImportMeta);
```
justinfagnani: use `path.replaceWith` instead?

packages/bundler/src/es6-rewriter.ts
usergenic: Added issue. https://github.com/Polymer/tools/issues/310

packages/bundler/src/es6-rewriter.ts
rictic: +1. Modifying documents returned by analyzer will lead to the worst kind of weird bugs with spooky action at a distance.

packages/bundler/src/es6-rewriter.ts
usergenic: Okay. That makes sense here. Replace'd

keanulee:
I'm seeing this transform present on the [es6-bundled](https://github.com/Polymer/pwa-starter-kit/blob/master/polymer.json#L32) build output which doesn't make since it uses `transformModulesToAmd` and makes `import` undefined/an invalid keyword in FF. Is this expected, or is there some other JS config I need in that build configuration @usergenic ?

```js
const bundledImportMeta = { ...import.meta,
  url: new URL('...', import.meta.url).href
};
```

usergenic:
@keanulee the fix was to pin a specific dependency (that was suddenly causing errors after a release a couple days ago) until babel's minify preset is more stable https://github.com/Polymer/tools/pull/380

PR #306, opened 2018-05-03
fix issue with routes being substrings but not root path requests

### LINKED ISSUE #302: [wct/polyserve] path truncation when one package is a prefix of another
Something is up in web-component-tester or polyserve where paths are being truncated in the case that one package is a prefix of another. This is manifesting as incorrect name-to-path conversion in some element tests.

To reproduce:

```bash
git clone git@github.com:PolymerElements/iron-a11y-keys
cd iron-a11y-keys
git checkout __auto_generated_3.0_preview
npm install
polymer version --version # 1.7.0-pre.16
polymer test --npm --module-resolution=node -p -l chrome
```

Results in 404s for files including `polymer-legacy.js`

Logging from within Analyzer's `resolve-specifier-node.ts` to see what's going on, I found a resolution happening where the `documentPath` has been truncated to `.../-behavior/...` which is probably then causing an incorrect relative path to be computed:

```
specifier: @polymer/polymer/polymer-legacy.js
documentPath: /Users/aomarks/git/polymer/iron-a11y-keys/-behavior/iron-a11y-keys-behavior.js
```

I would guess that something is doing a naive `startsWith` check somewhere, where it should be using `pathIsInside` (or doing nothing at all -- not sure why we would need to do any path truncation before we hit the name-to-path resolver). Not sure which package yet.

cc @e111077

e111077:
fixes #302

### DIFF
--- a/packages/polyserve/CHANGELOG.md
+++ b/packages/polyserve/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
 <!-- Add new, unreleased items here. -->
+* Fixed issue where the package name was only a substring of a request path was
+  being considered a root path request. e.g. a package of `@foo` with a request
+  path os `/@foo/bar` is considered a root path request, but a request path of
+  `/@foobar/baz` is not but was being considered one before.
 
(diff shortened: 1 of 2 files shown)


aomarks:
LGTM, can you update the CHANGELOG?

PR #307, opened 2018-05-03
Remove module index anchor from import.meta.url of top-level modules.

aomarks:


packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -215,7 +215,15 @@ function require(
 
     } else if (depSpec === 'meta') {
       numUnresolvedDeps--;
-      args.push({url: mod.url});
+      args.push({
+        // We append "#<script index>" to top-level scripts so that they have
+        // unique keys in the registry. We don't want to see that here. Note
+        // that the real base URL may legitimately have an anchor, so it's not
+        // safe to trim the anchor of non-top-level script URLs.
+        url: mod.isTopLevel === true ?
```
justinfagnani: style nit: I definitely prefer putting parens around the ternary condition.

PR #308, opened 2018-05-03
Add an integration test of our import.meta support.

rictic:
Will fail until https://github.com/Polymer/tools/pull/307 is merged


usergenic:
(master now has the import.meta stuff for bundler merged in.)

PR #309, opened 2018-05-03
Fix _transform in build to correspond to breakage from Typescript 2.8.3 Transform typings update

usergenic:
Separating this work out from #300 which is where I noticed the builds start failing.  I created a new branch from master to verify it was unrelated to that work.


usergenic:
Had a fix in https://github.com/PolymerLabs/css-slam/pull/19 that was the last piece of the puzzle.  Waiting for CI now; build should be fixed now...

rictic:
Isn't Transform is a typing in the node typings, not in TypeScript 2.8.3?

PR #312, opened 2018-05-03
Add nuke command, and don't allow CLI to be published without a shrinkwrap.

aomarks:


packages/cli/package.json
```diff
@@ -13,6 +13,7 @@
     "lint": "gulp lint",
     "build": "gulp build",
     "test": "gulp test lint",
+    "prepublishOnly": "ls npm-shrinkwrap.json",
```
rictic: ?

usergenic: I assumed this is here to exit non-0 when shrinkwrap file is missing.

aomarks: Correct. If you try to publish CLI but forgot to make a shrinkwrap, then it will fail with `ls: npm-shrinkwrap.json: No such file or directory`

Issue #295, opened 2018-05-02
[bundler] Rollup issue with module resolution


usergenic:
The repro was to clone https://github.com/material-components/material-components-web-components and try to bundle using one of the demo files as a fragment.  The result was that rollup complained about not being able to load a file, which resolved incorrectly I thing to a file URL... `file://something/something/demos/%40polymer/lit-element`

usergenic:
More specifically:
```
$ node ~/src/github.com/polymer/tools~active/packages/bundler/lib/bin/polymer-bundler.js --in-file demos/button.html --module-resolution node --inline-scripts
Error: Could not load file:///Users/usergenic/src/github.com/material-components/material-components-web-components/demos/%40polymer/lit-element/lit-element.js (imported by *bundle*): ENOENT: no such file or directory, open 'file:///Users/usergenic/src/github.com/material-components/material-components-web-components/demos/%40polymer/lit-element/lit-element.js'
    at /Users/usergenic/src/github.com/polymer/tools~active/packages/bundler/node_modules/rollup/dist/rollup.js:21430:19
    at <anonymous>
```

PR #318, opened 2018-05-03
[bundler] Added support for inline script type=module to use bare module specifiers

usergenic:
Fixes https://github.com/Polymer/tools/issues/295 which translates roughly: bare module specifiers in inline module scripts in html were not being connected to the Analyzer's JavaScriptImports and were being treated as local path names.  This code updates the rollup code to use the Document of the inline script to get the JavaScriptImports from it, operating the same way that subsequently imported files were treated and resolving bare name specifiers correctly.


packages/bundler/src/test/html-bundler_test.ts
```diff
@@ -33,6 +31,32 @@ const stripSpace = (html: string): string =>
...
+  test('inline es6 modules with node module-resolution', async () => {
+    const root = 'test/html/inline-es6-modules';
+    const analyzer = new Analyzer({
+      urlResolver: new FsUrlResolver(root),
+      urlLoader: new FsUrlLoader(root),
+      moduleResolution: 'node'
+    });
+    const bundler = new Bundler({analyzer});
+    const usesNpmPackageUrl = analyzer.resolveUrl('uses-npm-package.html')!;
+    const somePackageJsonUrl =
```
rictic: Unused?

packages/bundler/src/test/html-bundler_test.ts
```diff
@@ -33,6 +31,32 @@ const stripSpace = (html: string): string =>
...
 
+  test('inline es6 modules with node module-resolution', async () => {
+    const root = 'test/html/inline-es6-modules';
+    const analyzer = new Analyzer({
+      urlResolver: new FsUrlResolver(root),
+      urlLoader: new FsUrlLoader(root),
+      moduleResolution: 'node'
+    });
+    const bundler = new Bundler({analyzer});
+    const usesNpmPackageUrl = analyzer.resolveUrl('uses-npm-package.html')!;
```
rictic: For clarity, I think it'd be better to talk about using node module resolution rather than using npm packages, in these variable and file names. Node module resolution is the terminology used in the rest of the system for this

rictic:
Two recommendations in the organization of the test code, otherwise LGTM

packages/bundler/src/test/html-bundler_test.ts
usergenic: Unused and now removed. Thanks!

packages/bundler/src/test/html-bundler_test.ts
usergenic: Renamed to `import-using-node-module-resolution.html`

Issue #131, opened 2018-04-13
[polymer-build] [transformModulesToAmd] Could not resolve module specifier "require"


atomikolex:
When running  CLI 1.7.0-pre.10 to build 
`{ 
      "name": "bundled-amd", 
      "bundle": { 
        "inlineScripts": false 
      }, 
      "js": { 
        "transformModulesToAmd": true 
       } 
    } ` 

I am getting: Could not resolve module specifier "require" in file ... and
[BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.

Not using inline module scripts or import * syntax.

PR #320, opened 2018-05-03
Break out more shop tests.

rictic:
Adds tests of the source with a few different polyserve definitions.

Also breaks out each of the individual build configuration tests.

This isn't bad performance-wise, because the vast majority of the cost is in `npm install` and `polymer build` which is not repeated.

Also removes some unnecessary and distracting logging statements.

### DIFF
--- a/packages/build/src/babel-plugin-bare-specifiers.ts
+++ b/packages/build/src/babel-plugin-bare-specifiers.ts
@@ -94,12 +94,7 @@ const maybeResolve = (
     return resolve(specifier, filePath, componentInfo);
   } catch (e) {
     if (!isPathSpecifier(specifier)) {
-      // Don't warn if the specifier was already a path, even though we do
-      // resolve paths, because maybe the user is serving it some other
-      // way.
-      console.warn(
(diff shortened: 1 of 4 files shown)


packages/build/src/babel-plugin-bare-specifiers.ts
```diff
@@ -94,12 +94,7 @@ const maybeResolve = (
     return resolve(specifier, filePath, componentInfo);
   } catch (e) {
     if (!isPathSpecifier(specifier)) {
-      // Don't warn if the specifier was already a path, even though we do
-      // resolve paths, because maybe the user is serving it some other
-      // way.
-      console.warn(
```
aomarks: Would prefer to keep this warning since an unresolved module could be bad. The missing "require" (and now "meta") warnings we should def fix (https://github.com/Polymer/tools/issues/131) since everyone will see them and they are not bad.

rictic: +1, reverted. I think this now fixes #131

PR #322, opened 2018-05-04
 Change the default value of --module-resolution to "node"

justinfagnani:


rictic:
One random test after another keeps failing. This PR was passing before I rebased it, and I the tests which are failing and passing keep changing, so I'm confident that this is a flaky test issue. Merging.

PR #327, opened 2018-05-04
Add CHANGELOG entries for new compile target options.

aomarks:
These were forgotten earlier, but are pretty important.


aomarks:
> How are these targets specified? We should be more specific I think

Clarified, and also added the entry to polymer-project-config.

PR #333, opened 2018-05-04
Fix WCT under Windows

aomarks:
Fix node package file resolution under Windows. We were not accounting for Windows style paths. Also simplify the implementation a bit, and rename some test fixture files so that new implementation can really find the files it is resolving.

Fixes https://github.com/Polymer/web-component-tester/issues/694


packages/web-component-tester/runner/config.ts
```diff
@@ -175,43 +169,33 @@ export function truncatePathToDir(directory: string, pathName: string): string|
...
-        path.posix.relative(nodeModulesDir, absoluteBasePath);
-
-    resolvedEntrypoints.push(
-        path.posix.join(relativeBasePath, npmPackage.jsEntrypoint));
+    const absoluteFilepath = resolve.sync(
+        path.join(npmPackage.name, npmPackage.jsEntrypoint),
+        {basedir: wctPackageRoot});
+    const relativeFilepath = path.relative(rootNodeModules, absoluteFilepath);
+    // Convert Windows filesystem paths.
+    const relativeUrl = relativeFilepath.replace(/\\/g, '/');
```
rictic: We should do this only on Windows

aomarks: Done.

tomloake:
Excuse my naivety and please shout if there's a central place to check this stuff but does anyone know when this will be released?

aomarks:
Today!

Issue #332, opened 2018-05-04
[bundler] Weird failure to inline of script type=module results in empty inline script with blank src attribute


usergenic:
input:
```
<script type="module" src="../node_modules/@material/mwc-fab/mwc-fab.js"></script>
```
expected:
```
<script type="module">
// rolled-up contents of ../node_modules/@material/mwc-fab/mwc-fab.js
</script>
```
actual:
```
<script type="module" src=""></script>
```

Repro is:
```
git clone https://github.com/material-components/material-components-web-components
cd material-components-web-components
git checkout update-deps
npm install
lerna bootstrap ... something something
touch index.html
polymer build --bundle --entrypoint demos/fab.html
```

usergenic:
First thing to note is I forgot to include `--module-resolution node` in the invocation.  However, this doesn't seem to fix the problem when running `polymer build`.

```
polymer build --bundle --entrypoint demos/fab.html --module-resolution node --npm
```
Basically produces the same output.

However, using polymer-bundler CLI directly doesn't have this problem.  WTFBBQ:
```
polymer-bundler --module-resolution node demos/fab.html --out-dir build --inline-scripts
```
TOTALLY WORKS WTF

usergenic:
oh wait totally wrong it doesnt work.  if you leave off `--inline-scripts` it puts the mwc-fab.js into shared_bundle_2.js and links to that and things are okay.  it still is failing to inline for some reason.

so the good news is this is not a weirdo cli problem but a problem in bundler inlining code.  fixing.

usergenic:
Yay found the cause and fixed it in #334, currently in review.

PR #334, opened 2018-05-05
[bundler] Fixed the problem where we totally failed at inlining external module script tags.

usergenic:
Fixes https://github.com/Polymer/tools/issues/332 which was due to a misunderstanding of what `_inlineModuleScripts` was doing (it wasn't inlining module scripts, it was rolling up scripts that were already inline) and so external script type module tags were being ruined by having their attributes set to relative path to the bundle, which is an empty string.

This is fixed. 

Inlining strategy is basically creating a script tag with a single import statement to the thing to import and letting existing rollup code take care of the rest.

### DIFF
--- a/packages/build/src/bundle.ts
+++ b/packages/build/src/bundle.ts
@@ -152,18 +152,20 @@ export class BuildBundler extends AsyncTransformStream<File, File> {
    */
   private _unmapBundledFiles(manifest: BundleManifest) {
     for (const {
+           files,
            inlinedHtmlImports,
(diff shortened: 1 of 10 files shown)


usergenic:
@rictic I added a build to the polymer.json for the modules fixture which demonstrates the particular case addressed by this PR's updates to the html-bundler.ts - I may add more module bundling cases next week in a different PR.

usergenic:
Just rebased on master for clarity.

Issue #335, opened 2018-05-05
When loading via AMD loader, module execution order can be unexpected


sorvell:
This causes flakiness in ShadyCSS style processing because that library has load order requirements that users must adhere to.

1. a test loads [polymer-legacy.js](https://github.com/Polymer/polymer/blob/3.x-test-green/test/unit/styling-only-with-template.html#L18)
2. polymer-legacy loads
   * [legacy-element-mixin](https://github.com/Polymer/polymer/blob/3.x-test-green/polymer-legacy.js#L11)
     * legacy-element-mixin loads [apply-shim as its first import](https://github.com/Polymer/polymer/blob/3.x-test-green/lib/legacy/legacy-element-mixin.js#L10)
   * [custom-style](https://github.com/Polymer/polymer/blob/3.x-test-green/polymer-legacy.js#L20)

Based on this, I expect `apply-shim` to load and execute before `custom-style`.

However, it does not (verified via logging and Edge element inspector):

![screen shot 2018-05-04 at 7 33 29 pm](https://user-images.githubusercontent.com/78891/39658979-66a6ba80-4fd3-11e8-9a0c-ead3352d75bc.png)

rictic:
I think what's happening is that we're not following the full es6 import ordering semantics. If you've got:

#### start.js
```js
import 'foo';
import 'baz';
```

#### foo.js
```js
import 'bar';
```


Then the execution order is defined as: `bar, foo, baz, start`, but I think our loader is racy and might run baz _first_. 

So to do this right we can't just run module bodies when all of their imports are present, we also need to obey ordering of their parents' dependencies, which can be extra tricky when cycles get involved.

I'll put together some tests to verify that this is what's happening.

Issue #325, opened 2018-05-04
[esm-amd-loader] Tests do not run on Sauce


aomarks:
As demonstrated by the fact that our build was green even though the tests were failing in IE. Need to figure out how to get them running in our Sauce Travis build.

aomarks:
Fixed in https://github.com/Polymer/tools/pull/441

PR #339, opened 2018-05-05
Enforce more loader ordering strictness

rictic:
I _think_ this fixes https://github.com/Polymer/tools/issues/335

Formalizes the state machine a bit more. I don't think the typings are exactly right, as we still need casts in a number of places. Suggestions very welcome.

This will regress performance because while we're still able to load code from the network as fast as we were before, we're more constrained in when we can execute it, so the CPU will be idle more as the page loads.

We need more tests here, as there's a lot of subtle interactions and races possible here.

@sorvell would you mind trying this branch out with polymer core to see if it resolves the issue you were seeing?


packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,310 @@ interface Window {
...
+type NormalizedUrl = string&{_normalized: never};
+
+(function() {
+
+/**
+ * Describes the state machine for loading a single module.
+ *
+ * At runtime, these are integers that are inlined in their use sites.
+ */
+const enum StateEnum {
```
justinfagnani: I really don't like TS enums - they're one of the parts of TS that's not just a type-system superset of JS.

Given the great support for string literal and union types, I'd rather see a pattern like:

```ts
type State = 'Initialized'|'Loading'...;

/**
 * The initial state
 */
const initialized = 'Initialized' as 'Initialized';
```

This also lets you create other groups of states, like:

```ts
type WaitingState = 'WaitingOnEarlierScripts'|'WaitingOnDeps';
```

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,310 @@ interface Window {
...
+   * Args that we will pass into to moduleBody.
+   */
+  args: Array<{}>;
+  /**
+   * The body of the module, which is executed after its dependencies are
+   * loaded.
+   *
+   * In AMD/Commonjs terminology, this is the factory function.
+   */
+  moduleBody: undefined|Function;
```
justinfagnani: I'd rather name it as the factory then

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,310 @@ interface Window {
...
+}
+type ModuleStateData =
+    Initialized|Loading|WaitingOnEarlierScripts|WaitingOnDeps|Executed|Failed;
+
+interface Module<SD extends ModuleStateData = ModuleStateData> {
   url: NormalizedUrl;
   urlBase: NormalizedUrl;
   exports: {[id: string]: {}};
-  /** This is a top-level module. */
+  stateData: SD;
```
justinfagnani: Is there a reason why the state data is separate? Is it to allow the factory function to be freed? I feel like it'd be simpler in code to just have a module keep it's factory and deps.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,310 @@ interface Window {
...
-  /** This module hasn't started loading its script yet. */
-  needsLoad: boolean;
-  /** All dependencies are resolved and the factory has run. */
-  resolved: boolean;
-  error: Error|undefined;
-  /** Callbacks from dependents waiting for this module to resolve. */
-  onResolve: Array<() => void>;
-  onError: Array<ErrorCallback>;
+  /**
+   * Callbacks that are called exactly once, for the next time the module
```
justinfagnani: What kind of callbacks? Is there a named type that can describe this here and be searchable in the rest of the module?

rictic: We only use this in two places, and since this is a zero argument function that returns nothing I'd say just look for `onNextStateChange`.

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: The primary reason was that when I was first writing out the state machine, I was spending a lot of documentation on when data was available, when I realized this was the job of the type system. Saving some memory was just a small bonus.

I'm not super happy about this implementation though, because the type of Module isn't narrowed when we check on its stateData's type. Gonna play with some other options.

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: We're using ES standard terminology in other places (e.g. we call them dependency specifiers, not ids), which I think is desirable.

I also think it's really weird to refer to a function that's only called once and doesn't return anything as a factory.

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: That's normally what I would lean towards too, but this form keeps about the same ergonomics and legibility and compiles down to a single byte. If we want to convert this to JS we could maintain this by just declaring a const for each enum value, and we could count in them being inlined as they are now.

aomarks:
I think it's possible to deadlock here even when your explicit dependencies are still a DAG, because cycles can be introduced implicitly by the ordering constraint. For example, if in the test you make `foo` depend on `bar` (which does not introduce an explicit cycle), then the test times out: https://github.com/Polymer/tools/commit/568697c92b51782dffff715d1b6fdbf5c077b010.

```
start -> foo -> baz
  |       |
  |       v
  +----> bar
```

rictic:
@aomarks I pulled that locally, and it does initially look like a deadlook, but it's actually the assertion failing, which is caught and hidden because we now execution module bodies in a try/catch so that we can propagate the failure.

I think should also rethrow exceptions when the module isn't being loaded in a dynamic import.

sorvell:
The order is still incorrect for top level modules in HTML. I see this order in the network tab:
  1. `polymer-legacy.js` 
  1. `polymer-element.js`

This is unexepcted since all the the transitive dependencies of `polymer-legacy.js` should load before the second script type module.

screen shot 2018-05-05 at 3 30 37 pm

kevinpschaaf:
@aomarks @rictic For scripts in the main document, I thought we were injecting dependencies on the previous script to maintain "defer" ordering semantics of module scripts.  Based on that screenshot above, looks like they're unordered and will just kick off loading their depndencies in parallel, which seems bad.

rictic:
We now do the native ESM thing here. We _load_ scripts whenever we first find out about them (this is unchanged from before this PR), but we execute the body of the script (the second argument to `define`) in strict dependency order.

This lets us still get good performance, but it does mean that you can't rely on the order of the script tags or the network tab for execution order.

rictic:
I believe this is ready for review. @sorvell has confirmed that it fixes the original issue, and I've run the loader's tests on Safari and Edge locally, in addition to the Chrome and Firefox tests on travis.

Fixes #335

TimvdLippe:
What is weird is that the tests did pass on Edge 17 when I tried it out and they are worked in https://travis-ci.org/Polymer/polymer/builds/375534021 Still, Safari 10 and 11 are broken there, so I think this change is still necessary.

rictic:
That's not super surprising. For most projects this change won't have any effect, and for Polymer core I think there's just one place where the order matters. This change ensured that it goes in the standard order, but you could luck into the standard order with the loader at master.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,331 @@ interface Window {
...
+    if (onResolve) {
+      onResolve();
+    }
+    return;
+  }
+  const nextDep = deps[0];
+  switch (nextDep.stateData.state) {
+    // Cases where the dep needs us to push it forward.
+    case StateEnum.Initialized:
+      load(nextDep as Module<Initialized>);
```
justinfagnani: this isn't clear to me that loads aren't serialized. Why only load the nextDep?

rictic: Honestly I don't _think_ this line of code is ever executed, as `beginWaitingOnEarlierScripts` should have ensured that all of our dependencies have made it at least to `Loading`. The tests all pass with this line replaced with `throw new Error()`, this is just a belt and suspenders sort of check. Maybe unnecessary? WDYT

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+    } else {
+      // The script did not make a call to define(), otherwise the global
+      // callback would have been set. That's fine, we can resolve immediately
+      // because we don't have any dependencies, by definition.
+      deps = [];
+      moduleBody = undefined;
+    }
+    beginWaitingOnEarlierScripts(mutatedModule, deps, moduleBody);
+    try {
+      document.head.removeChild(script);
```
justinfagnani: do we care about removing scripts? why?

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+    if (pendingDefine !== undefined) {
+      [deps, moduleBody] = pendingDefine();
+    } else {
+      // The script did not make a call to define(), otherwise the global
+      // callback would have been set. That's fine, we can resolve immediately
+      // because we don't have any dependencies, by definition.
+      deps = [];
+      moduleBody = undefined;
+    }
+    beginWaitingOnEarlierScripts(mutatedModule, deps, moduleBody);
```
justinfagnani: why do we get to start waiting on earlier scripts now? Seems like we should be waiting on deps first. Is that name really descriptive of what it's doing?

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -87,26 +394,58 @@ window.define = function(deps: string[], factory?: ResolveCallback) {
...
       mod.isTopLevel = true;
-      mod.needsLoad = false;
-
-      if (previousTopLevelUrl !== undefined) {
+      const predecessor = previousTopLevelUrl;
+      previousTopLevelUrl = url;
+      const waitingModule = beginWaitingOnEarlierScripts(mod, deps, factory);
+      const nextStep = () => {
+        beginWaitingOnDeps(waitingModule);
+        function failureShouldThrow(module: Module) {
```
aomarks: could this logic move into the fail function? there is already a `topLevel` property you can use to know that it needs to throw.

either way, add a little comment to make it clear this is the special handling for top level

packages/esm-amd-loader/src/esm-amd-loader.ts
aomarks: I assumed this was more about size over the wire?

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+  return stateTransition(mod, {state: StateEnum.Failed, error});
+}
+
+/**
+ * Transition the given module to the given state.
+ *
+ * Does not do any checking that the transition is legal.
+ * Calls onNextStateChange callbacks.
+ */
+function stateTransition<SD extends ModuleStateData>(
```
aomarks: Move this up higher? Feels like this should be close to the state enum since it's so general to the state machine.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+    return;
+  }
+
+  if (nextDep.stateData.state === StateEnum.WaitingOnDeps) {
+    if (debugging) {
+      console.log(`Cycle detected importing ${nextDep.url}`);
+    }
+    // Do not wait on the dep that introduces a cycle, press on as though it
+    // were not there.
+    waitOnDeps(deps, onResolve, onError);
```
aomarks: should there be a `return` here? seems like you still end up waiting on the dep right now.

packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -159,12 +211,25 @@ suite('top-level modules', () => {
...
+
+    test('can fail without blocking the next one', (done) => {
+      window.uncaughtErrorFilter = (e) => {
+        // We only expect this error once.
+        window.uncaughtErrorFilter = undefined;
+        // Look for a failure to fetch the file that does not exist.
+        return /Failed to fetch .*static\/y\/not-found\.js/.test(
+            e.error.message);
+      };
+      // We order top-level modules by injecting dependencies between them.
```
aomarks: Not true anymore.

packages/esm-amd-loader/CHANGELOG.md
```diff
@@ -5,7 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- # Unreleased -->
+# Unreleased
+* Fix bug where the ordering of module execution was not strict enough.
+  Standard ES modules are loaded as they're encountered, but they execute
+  strictly in the order that they're first imported.
```
aomarks: Mention cycles support?

packages/esm-amd-loader/README.md
```diff
@@ -2,7 +2,7 @@
 
 # @polymer/esm-amd-loader
 
-A JavaScript library which loads AMD-style modules in the browser in 0.9 KB.
+A JavaScript library which loads AMD-style modules in the browser in 1.2 KB.
```
aomarks: Size is also mentioned below. Also you might want to update some of the documentation with the renamed things, e.g. factory is gone.

packages/esm-amd-loader/package.json
```diff
@@ -15,6 +15,7 @@
     "lint": "tslint --project . --format stylish",
     "build": "npm run clean && tsc && npm run minify",
     "minify": "babel-minify lib/esm-amd-loader.js --outFile=lib/esm-amd-loader.min.js",
+    "get-size": "npm run build && cat lib/esm-amd-loader.min.js | gzip -9 > esm-amd-loader.min.js.gz && ls -a -l -h esm-amd-loader.min.js.gz && rm esm-amd-loader.min.js.gz",
```
aomarks: `npm run build && gzip -c -9 lib/esm-amd-loader.min.js | wc -c`

packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -256,14 +321,75 @@ suite('meta.url', () => {
...
+  // suite.html#0  -> start-one
+  // suite.html#1  -> start-two
+  // start-one -> [a, e]
+  // start-two -> [a, g, h]
+  // a -> [b, e]
+  // b -> [c, d]
+  // e -> [f, g]
+  // h -> [i, j, k]
+  const testName =
+      'with a deep and more complicated tree of deps, they still execute in import order according to spec';
```
justinfagnani: what specifically is the test trying to cover with a larger dep tree? Is it just trying to trigger the out-of-order execution due to one chain of deps being longer? Seems like this would be more deterministic and require a small graph if we just setup a server to delay serving of a module earlier in the total order and make sure a module later in the order that doesn't depend on it doesn't run first.

justinfagnani:
> Honestly I don't think this line of code is ever executed, as beginWaitingOnEarlierScripts should have ensured that all of our dependencies have made it at least to Loading. The tests all pass with this line replaced with throw new Error(), this is just a belt and suspenders sort of check. Maybe unnecessary? WDYT

I like asserting your assumptions and failing quick, rather than trying to do something safe if they fail. The safe thing can too easily be carried far into the future complicating the code.

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: Many DOM algorithms are O(tags in document)

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: +1, done

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+}
+
+function waitOnDeps(
+    deps: Module[],
+    onResolve: ResolveCallback|undefined,
+    onError: ErrorCallback|undefined): void {
+  const nextDep = deps.shift();
+  if (nextDep === undefined) {
+    // all deps resolved
+    if (onResolve) {
```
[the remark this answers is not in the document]
rictic: This could be from a user code `require()` call, so I thought a falsy check would work better.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -18,42 +18,349 @@ interface Window {
...
+    for (const callback of callbacks) {
+      callback();
+    }
+  }
+  return mutatedModule;
+}
+
+function waitOnDeps(
+    deps: Module[],
+    onResolve: ResolveCallback|undefined,
```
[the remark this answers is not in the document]
rictic: +1

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: Yep, that's a bug!

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -73,10 +362,10 @@ window.define = function(deps: string[], factory?: ResolveCallback) {
   // fired by this <script>, which will be handled by the loading script,
   // which will invoke the callback with our module object.
   let defined = false;
-  pendingDefine = (mod) => {
+  pendingDefine = () => {
     defined = true;
     pendingDefine = undefined;
-    define(mod, deps, factory);
+    return [deps, factory];
```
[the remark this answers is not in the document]
rictic: Consistentified

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -87,26 +376,58 @@ window.define = function(deps: string[], factory?: ResolveCallback) {
     if (defined === false) {
       pendingDefine = undefined;
       const url = baseUrl + '#' + topLevelScriptIdx++ as NormalizedUrl;
-      const mod = getModule(url);
+      // It's actually Initialized, but we're skipping over the Loading
+      // state, because this is a top level document and it's already loaded.
+      const mod = getModule(url) as Module<Loading>;
 
       // Top-level scripts are already loaded.
```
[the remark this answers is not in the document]
rictic: Done

packages/esm-amd-loader/src/esm-amd-loader.ts
rictic: Oh good point! Yeah that's much more elegant.

packages/esm-amd-loader/test/src/suite.ts
rictic: Updated comment

packages/esm-amd-loader/test/src/suite.ts
rictic: Yeah, this test name isn't good. The key thing being tested here but not in the simpler test is that the modules are ordered globally, not locally. i.e. a module isn't ready to run just because it's the first dependency in a define call, it also matters what order its parent runs in.

Updated the test name.

rictic:
All comments addressed, PTAL

packages/esm-amd-loader/package.json
aomarks: Ah ok. Can kill some of the cats and pipes still: `npm run build && gzip -9 esm-amd-loader.min.js && ls -l -h esm-amd-loader.min.js.gz && rm esm-amd-loader.min.js.gz`

packages/esm-amd-loader/src/esm-amd-loader.ts
aomarks: Add a comment? RequireJS doesn't do this FYI.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -13,55 +13,377 @@
...
+  /**
+   * The body of the module, which is executed after its dependencies have
+   * executed.
+   *
+   * In AMD/Commonjs terminology, this is the factory function.
+   */
+  readonly moduleBody: undefined|Function;
+}
+
+interface ModuleG<State extends keyof StateDataMap = keyof StateDataMap> {
```
aomarks: What does the G stand for?

packages/esm-amd-loader/package.json
rictic: That deletes esm-amd-loader.min.js which causes tests to fail (which is Good and Correct), but that's why I went with the way more roundabout script

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -13,55 +13,377 @@
...
+  }
+  return stateTransition(module, StateEnum.Executed, undefined);
+}
+
+/**
+ * Called when a module has failed to load, either becuase its script errored,
+ * or because one of its transitive dependencies errored.
+ */
+function fail(mod: Module, error: Error) {
+  if (mod.isTopLevel) {
```
aomarks: === true

packages/esm-amd-loader/package.json
aomarks: Oh, there is the `-k` or `--keep` flag to gzip which does the same thing. Whatever though.

packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -87,26 +409,41 @@ window.define = function(deps: string[], factory?: ResolveCallback) {
...
-
-      // Top-level scripts are already loaded.
+      // It's actually Initialized, but we're skipping over the Loading
+      // state, because this is a top level document and it's already loaded.
+      const mod = getModule(url) as ModuleG<StateEnum.Loading>;
       mod.isTopLevel = true;
-      mod.needsLoad = false;
-
-      if (previousTopLevelUrl !== undefined) {
+      const predecessor = previousTopLevelUrl;
```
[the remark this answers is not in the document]
rictic: Nice, yeah

aomarks:
Manually verified tests pass in IE and Edge. I'll work on https://github.com/Polymer/tools/issues/325 so we dont have to verify manually next time.

PR #340, opened 2018-05-06
Remove @types/rollup from bundler

TimvdLippe:
These typings are now being shipped by Rollup itself: https://www.npmjs.com/package/@types/rollup

My `npm install` reordered some fields in the `package.json`. I assume this is new npm behavior :thinking: The content is still the same though. (For example `license` was duplicate specified)

### DIFF
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -4,7 +4,7 @@
   "description": "Process Web Components into one output file",
   "homepage": "https://github.com/Polymer/tools/tree/master/packages/bundler",
   "repository": "github:Polymer/tools",
-  "bugs": "https://github.com/Polymer/polymer-bundler/issues",
+  "bugs": "https://github.com/Polymer/tools/issues",
   "license": "BSD-3-Clause",
   "author": "The Polymer Project Authors",
   "main": "lib/bundler.js",
--- a/packages/bundler/package.json
+++ b/packages/bundler/package.json
@@ -3,13 +3,8 @@
   "version": "4.0.0-pre.7",
   "description": "Process Web Components into one output file",
   "homepage": "https://github.com/Polymer/tools/tree/master/packages/bundler",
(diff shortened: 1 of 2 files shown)


packages/bundler/package.json
```diff
@@ -3,8 +3,13 @@
   "version": "4.0.0-pre.7",
   "description": "Process Web Components into one output file",
   "homepage": "https://github.com/Polymer/tools/tree/master/packages/bundler",
-  "repository": "github:Polymer/tools",
-  "bugs": "https://github.com/Polymer/tools/issues",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/Polymer/polymer-bundler.git"
```
usergenic: This was actually changed to `"repository": "github:Polymer/tools"` on purpose.

packages/bundler/package.json
```diff
@@ -3,8 +3,13 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/Polymer/polymer-bundler/issues"
+  },
```
usergenic: Change back to `"bugs": "https://github.com/Polymer/tools/issues"`

TimvdLippe:
@usergenic These changes were done automatically by npm 6.0. I think that if I revert them, the next time it will apply the changes again?

usergenic:
I don't think so. I suspect this is symptomatic of moving to the monorepo from the separate bundler repo and residual something something in the matrix...  There's no reason to expect other than unaccounted-for-weirdness that this should happen.

TimvdLippe:
Ugh, I hate the GitHub UI. Have to do this tomorrow on my laptop to clean up this mes....

TimvdLippe:
Or actually, the last commite message is just bad, but the content is good. Apparently these repository and bugs fields were duplicate, so thats why my npm changed them 🤔

PR #342, opened 2018-05-06
Re-enable editor-service tests

### LINKED ISSUE #170: The definition finder editor-service tests are a bit flaky

TimvdLippe:
I was not able to let them fail locally, so let's see what CI says.

Fixes #170


rictic:
It's only very occasionally flaky. We might have turned on retries here though, in which case this should be fine.

TimvdLippe:
Superseded by #429

PR #344, opened 2018-05-07
[bundler] Tie up a few small loose ends in bundler code

usergenic:
- Removed an obsolete TODO about `export * from` which was previously fixed. 5fa64a1
- Use the already defined assertJsDocument() and avoid unnecessary cast. 7e9f6a5
- Generate unique identifier name for the bundledImportMeta stand-in variable for bundled import.meta transform. 6b907ae

### DIFF
--- a/packages/bundler/src/es6-rewriter.ts
+++ b/packages/bundler/src/es6-rewriter.ts
@@ -11,6 +11,7 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
+import generate from '@babel/generator';
 import traverse, {NodePath} from 'babel-traverse';
 import * as babel from 'babel-types';
 import * as clone from 'clone';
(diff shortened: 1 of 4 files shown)


packages/bundler/src/es6-rewriter.ts
```diff
@@ -485,21 +486,27 @@ export class Es6Rewriter {
...
-    // const __bundledImportMeta = {...import.meta, url: __bundledImportURL};
+    // ```javascript
+    // const bundledImportMeta = {
+    //    ...import.meta,
+    //    url: new URL(${ relativeUrl }, import.meta.url).href
+    // };
+    // ```
     // TODO(usergenic): Consider migrating this AST production mishmash into the
     // `ast` tagged template literal available like this:
     // https://github.com/Polymer/tools/blob/master/packages/build/src/babel-plugin-dynamic-import-amd.ts#L64
```
usergenic: I've got another PR coming for this.

usergenic:
In reviewing the integration test failure output in appveyor and travis it appears to be unrelated to this PR.  Merging.

PR #345, opened 2018-05-07
Include node_modules/webcomponentsjs/*.js in extraDependencies

justinfagnani:


packages/cli/src/polymer-cli.ts
```diff
@@ -132,6 +132,8 @@ export class PolymerCli {
         this.defaultConfigOptions.extraDependencies || [];
     this.defaultConfigOptions.extraDependencies.unshift(
         `bower_components${pathSeperator}webcomponentsjs${pathSeperator}*.js`);
+    this.defaultConfigOptions.extraDependencies.unshift(
+        `node_modules${pathSeperator}webcomponentsjs${pathSeperator}*.js`);
```
bicknellr: @keanulee mentioned in chat that the bundles have moved to a separate folder, so this might need to be something like `node_modules${pathSeperator}webcomponentsjs${pathSeperator}**${pathSeparator}*.js`?

aomarks (approved):

packages/cli/src/polymer-cli.ts
bicknellr: Then again, I don't know what our glob syntax is. :/

justinfagnani: Oh, your right, this changes on npm

abdonrd:
Exclude `gulpfile.js`? As PSK does.

packages/cli/src/polymer-cli.ts
```diff
@@ -131,7 +131,13 @@ export class PolymerCli {
...
         this.defaultConfigOptions.extraDependencies || [];
     this.defaultConfigOptions.extraDependencies.unshift(
-        `bower_components${pathSeperator}webcomponentsjs${pathSeperator}*.js`);
+        path.join('!bower_components', 'webcomponentsjs', 'gulpfile.js'));
+    this.defaultConfigOptions.extraDependencies.unshift(
+        path.join('bower_components', 'webcomponentsjs', '*.js'));
+    this.defaultConfigOptions.extraDependencies.unshift(path.join(
+        '!node_modules', '@webcomponents', 'webcomponentsjs', 'gulpfile.js'));
+    this.defaultConfigOptions.extraDependencies.unshift(
+        path.join('node_modules', '@webcomponents', 'webcomponentsjs', '*.js'));
```
bicknellr: This doesn't look like it goes into`/bundles`, does it?

justinfagnani: good point

abdonrd:
@justinfagnani @bicknellr what do you think about pushing this PR?

https://github.com/webcomponents/webcomponentsjs/pull/929

We could avoid these things in `polymer.json`, `sw-precache-config.js`, PSK, etc.

bicknellr:
webcomponents/webcomponentsjs#929 seems like a reasonable thing to do, but does `package.json`'s `files` field have any special meaning to the tools repo? Specifically, does adding something to `files` mean that it can be excluded from `polymer.json`'s `extraDependencies`?

keanulee:
Left a comment in support of https://github.com/webcomponents/webcomponentsjs/pull/929#issuecomment-388141918. According to npm docs, [the files field](https://docs.npmjs.com/files/package.json#files) should only have effect when installing as a dependency. Still probably needs to be in `extraDependencies` (since specific bundles won't be reached by analyzer), but the whole entry can be simplified to `"node_modules/@webcomponents/webcomponentsjs/**"` (as in [shop#3.0](https://github.com/Polymer/shop/blob/3.0/polymer.json)).

abdonrd:
@keanulee is right! :)

PR #346, opened 2018-05-07
[cli] Update dependencies in the built-in templates.

bicknellr:


rictic:
Unrelated note, but polymer lint should be working in the `@next` version of polymer cli

PR #347, opened 2018-05-08
Simplify Travis build matrix.

aomarks:
Simplify Travis config by using "include" to set the matrix directly.

Also scopes browser installation and xvfb to only those builds that need it.


.travis.yml
```diff
@@ -1,41 +1,52 @@
...
-  - env: TEST_COMMAND="npm run test:integration" SKIP_REMOTE_BROWSERS=true
-    node_js: '9'
-language: node_js
-node_js:
-- '8'
-- '9' # TODO(https://github.com/Polymer/tools/issues/280): update to node 10
+
+branches:
+  only:
+  - master
```
TimvdLippe: Doesn't this prevent any PR from being built?

.travis.yml
```diff
@@ -1,41 +1,52 @@
...
-matrix:
-  exclude:
-  - env: TEST_COMMAND="npm run test:integration" RUN_SAUCE_TESTS=true
-    node_js: '8'
-  - env: TEST_COMMAND="npm run test:integration" SKIP_REMOTE_BROWSERS=true
-    node_js: '9'
-language: node_js
-node_js:
-- '8'
-- '9' # TODO(https://github.com/Polymer/tools/issues/280): update to node 10
```
TimvdLippe: Maybe add this command just above the matrix so that we are tracking it still?

aomarks: Done.

.travis.yml
aomarks: By default you get two Travis builds for a PR: one that builds the feature branch directly (called "branch"), and another that makes a hypothetical merge commit on master (called "PR"). We decided that the merge build is what really matters, so we don't want to waste resources by also building the branch.

(BTW this change has already been here for a while, I just moved it in the file).

TimvdLippe: Ah that makes sense. Maybe add this as a comment for historical perspective to help our future selves? 😛

aomarks: Done.

aomarks:
> Wait, so we can express what we want it to do instead of a superset of all things it can do and then excluding what we don't want it to do? Where's the fun in that?!?!
> 
> LGTM

:D

PR #351, opened 2018-05-08
Handle errors early in analysis better.

rictic:
Previously, there were many places for error messages to get lost, in part because we weren't tracking errors in the type system. This moves all expected errors in AnalysisContext into the type system with the Result type.

### DIFF
--- a/packages/analyzer/src/core/analysis-context.ts
+++ b/packages/analyzer/src/core/analysis-context.ts
@@ -12,6 +12,7 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import {isCancel} from 'cancel-token';
 import * as path from 'path';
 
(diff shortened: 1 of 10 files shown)


packages/analyzer/src/core/analysis-context.ts
```diff
@@ -329,10 +343,23 @@ export class AnalysisContext {
                     this.resolver.resolve(parsedDoc.baseUrl, i.url, i)));
             this._cache.dependencyGraph.addDocument(resolvedUrl, importUrls);
 
-            return scannedDocument;
+            return {successful: true, value: scannedDocument};
           } catch (e) {
-            this._cache.dependencyGraph.rejectDocument(resolvedUrl, e);
-            throw e;
+            const message = (e && e.message) || `Uknown error during scan.`;
```
aomarks: Unknown

rictic: Good catch!

PR #354, opened 2018-05-08
Allow the user to override parsing of some files.

### LINKED ISSUE #353: API to let the user override the expected content type of files
When we pipe through file types through from imports, we should also pipe in a way for the user to customize as well. A function that takes a resolved url and returns a content type should be sufficient.

Requested by the inimitable @ChadKillingsworth

rictic:
Fixes https://github.com/Polymer/tools/issues/353


aomarks:
Update CHANGELOG

rictic:
Updated!

aomarks (approved):

usergenic:
So currently, this fix results in one of those very Analyzer-y headscratcher bugs:

Screen Shot 2019-07-03 at 3 33 37 PM

Issue #357, opened 2018-05-09
[cli] polymer 3 template tests fail


usergenic:
Same failure for both polymer 3.x application and element templates.  Repro steps:

```
$ polymer init
(choose either polymer 3.x template)
...
$ polymer test
Installing and starting Selenium server for local browsers
Selenium server running on port 61127
[BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.
[BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.               safari 11.1 ()
safari 11.1              Beginning tests via http://localhost:8081/components/poly3-element/generated-index.html?cli_browser_id=3
[BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.
chrome 66                Beginning tests via http://localhost:8081/components/poly3-element/generated-index.html?cli_browser_id=1
chrome 68                Beginning tests via http://localhost:8081/components/poly3-element/generated-index.html?cli_browser_id=0
safari failed to maximize
chrome failed to maximize
chrome failed to maximize
firefox 59               Beginning tests via http://localhost:8081/components/poly3-element/generated-index.html?cli_browser_id=2
safari 11.1              ✖ test/poly3-element_test.html » poly3-element » instantiating the element with default properties works

  expected undefined to equal 'poly3-element'
    <unknown> at   AssertionError at /components/chai/chai.js:5260:0
    <unknown> at           assert at /components/chai/chai.js:259:0
    <unknown> at            equal at /components/chai/chai.js:2006:0
    <unknown> at        <unknown> at poly3-element_test.html:29:0

safari 11.1              ✖ test/poly3-element_test.html » poly3-element » setting a property on the element works

  null is not an object (evaluating 'elementShadowRoot.querySelector')
    <unknown> at   <unknown> at poly3-element_test.html:39:0

safari 11.1              Tests failed: 2 failed tests
firefox 59               ✖ test/poly3-element_test.html » poly3-element » instantiating the element with default properties works

  the string "console.error: Error stamping [object HTMLTemplateElement] console.error: template getter must return HTMLTemplateElement" was thrown, throw an Error :)
    <unknown> at   <unknown> at /components/mocha/mocha.js:1248:0

firefox 59               ✖ test/poly3-element_test.html » poly3-element » setting a property on the element works

  elementShadowRoot is null
    <unknown> at   <unknown> at poly3-element_test.html:80:0

[...]

PR #358, opened 2018-05-09
[cli] Fixed polymer 3.x templates and updated CHANGELOG to prepare for CLI release.

usergenic:
Fixes https://github.com/Polymer/tools/issues/357 and when we release, will reduce download size for users by 250mb.


usergenic:
$ npm install polymer-cli
1.7.0 : 553M 
1.7.1 : 286M

PR #360, opened 2018-05-10
[cli] Specify dependencies explicitly for GitHub generators in init.

### LINKED ISSUE #341: Is bower still needed in Polymer Starter Kit 4?
`polymer-cli` still run `bower install` in latest Polymer Starter Kit 4.

bicknellr:
- Pass dependency sources explicitly to the Yeoman generator that wraps GitHub projects.
- Bump PSK4 to the release version. (This should already be used during testing due to the hat.)
- Add an explicit version for the Shop generator tests. (^2.0.0)

Fixes #341. cc @motss

### DIFF
--- a/packages/cli/src/test/integration/integration_test.ts
+++ b/packages/cli/src/test/integration/integration_test.ts
@@ -164,7 +164,12 @@ suite('integration tests', function() {
       const ShopGenerator = createGithubGenerator({
         owner: 'Polymer',
         repo: 'shop',
+        semverRange: '^2.0.0',
         githubToken,
+        installDependencies: {
+          bower: true,
+          npm: false,
+        },
       });
 
       const dir = await runGenerator(ShopGenerator).toPromise();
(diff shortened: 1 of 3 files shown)


packages/cli/src/test/integration/integration_test.ts
```diff
@@ -503,8 +516,16 @@ suite('polymer shop', function() {
...
         this.timeout(2 * 60 * 1000);
-        const ShopGenerator = createGithubGenerator(
-            {owner: 'Polymer', repo: 'shop', githubToken, branch: '3.0'});
+        const ShopGenerator = createGithubGenerator({
+          owner: 'Polymer',
+          repo: 'shop',
+          githubToken,
+          branch: '3.0',
+          installDependencies: {
+            bower: false,
```
aomarks: do you need the explicit bower:false? why not just npm:true?

bicknellr: npm and bower are always run unless you disable them: https://github.com/yeoman/generator/blob/v1.0.1/lib/actions/install.js#L129

aomarks: ah cool

bicknellr:
And again, this time a different test.

motss:
@bicknellr 

Cool. 👍 LGTM. Thanks for the fix!

PR #362, opened 2018-05-10
[all] Make build green again (analyzer/bundler/cli/polyserve all updated)

usergenic:
Fixes https://github.com/Polymer/pwa-starter-kit/issues/85


packages/polyserve/src/start_server.ts
```diff
@@ -555,3 +558,13 @@ async function tryStartWithPort(
     });
   });
 }
+
+// TODO(usergenic): Something changed in the typings of net.Server.address() in that
```
aomarks: This was https://github.com/DefinitelyTyped/DefinitelyTyped/pull/25597. It is apparently correct that net.Server can return a string, but only when it's over pipes or Unix domain sockets, according to https://nodejs.org/api/net.html#net_server_address. I wonder if it's possible start an HTTP server over either of those. If not maybe the HTTP Server class type should be updated to override address() to say that it can only return the info object.

Flarna: @aomarks I have seen HTTP servers connected via a Windows Pipe on Microsoft azure web apps.

packages/cli/src/test/unit/init/init_test.ts
```diff
@@ -240,7 +238,7 @@ suite('init', () => {
             foo: 'TEST',
           }));
       helpers.registerDependencies(yeomanEnv, [[
-                                     helpers.createDummyGenerator(),
+                                     <any>function() {},
```
usergenic: yeoman-test installs yeoman-generator 2.x which is now incompatible with its `createDummyGenerator()` function because of the absence of Generator.extend() static method.

Simplest thing here to do was drop a meaningless stub function into the test register dependencies call.

Issue #227, opened 2018-04-25
[wct] Update Mocha


merlinnot:
We should update Mocha version, since it resolves multiple bugs and introduces some new features. It also contains some braking changes, so it requires a major version bump of WCT. Please consult Mocha changelog to learn more.

jlaamanen:
I would like to add that Mocha versions prior to v4 (WCT has ^3.4.2) use growl version 1.9.2, which has a critical vulnerability according to Node security: https://nodesecurity.io/advisories/146. Because of this, `npm audit` fails for projects that have web-component-tester listed as a dependency.

PR #364, opened 2018-05-10
Update packages to resolve many security audit warnings.

rictic:


packages/cli/src/polymer-cli.ts
```diff
@@ -195,7 +195,9 @@ export class PolymerCli {
         {args: commandArgs});
 
     const commandDefinitions = mergeArguments([command.args, globalArguments]);
-    const commandOptionsRaw = commandLineArgs(commandDefinitions, commandArgs);
+    console.log(`commandLineArgs: ${JSON.stringify(commandLineArgs, null, 2
```
TimvdLippe: Leftover debug statement I suppose?

packages/polyserve/src/test/cli_test.ts
```diff
@@ -58,6 +58,8 @@ suite('cli', () => {
       }
     } finally {
       unintercept();
+      console.log('stdout from failed cli invocation:');
+      console.log(stdout);
```
TimvdLippe: Are these leftover as well?

rictic: This are intended, the other ones aren't

packages/cli/src/polymer-cli.ts
rictic: good catch!

TimvdLippe:
We are upgrading to Mocha 5 here, but we are not updating `wct` as well. Reference issue: #227

merlinnot:
So... #377? 😸

PR #366, opened 2018-05-10
Fix incorrect module specifiers in polymer-3-element init template.

### LINKED ISSUE #365: polymer serve in a polymer-3-element won't work
After creating a new custom element using the Polymer CLI's polymer-3-element template, using polymer serve will throw an error on the browser console and the custom element won't render.

Steps to reproduce:
1. Install Polymer CLI 1.7.1
1. create a new web component using the polymer init command
1. use the polymer-3-element
1. run polymer serve
1. the following error appear on the console and the web component won't render:
screen shot 2018-05-10 at 21 53 17

aomarks:
We always need to use bare module specifiers, not relative URLs, when we refer to our dependencies. That means using inline scripts with `import` statements, and not external script tags.  In this case it caused some Polymer modules to be loaded twice, once from "/components/my-element/node_modules/@polymer/..." and again from "/components/node_modules/@polymer/...".

Also bump the minimum version of iron-demo-helpers, since I was getting an old version of prism-element that had a missing "prism-theme-default.js".

Fixes #365

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
 
-<!-- ## Unreleased -->
+## Unreleased
+* Fix bug in `polymer-3-element` init template where `polymer/dom-module.js`
+  could be loaded twice when serving from the polyserve `/components/`
+  directory.
 <!-- Add new, unreleased items here. -->
 
 ## v1.7.1 [05-09-2018]


bicknellr:
> We always need to use bare module specifiers, not relative URLs, when we refer to our dependencies.

But we can't do this for classic scripts, right? https://github.com/Polymer/tools/pull/366/files#diff-59d2389829a2bc107f1def84055d9db2R9 Didn't we decide that we don't want to serve from `/components/` anymore?

bicknellr:
Oh, is this causing a problem when testing through WCT?

aomarks:
> Oh, is this causing a problem when testing through WCT?

See #365. If you init `polymer-3-element`, then run polymer serve, then visit the URL that it tells you to (which still includes the /components/ directory), then you get errors with double-registration of dom-module.

> But we can't do this for classic scripts, right? https://github.com/Polymer/tools/pull/366/files#diff-59d2389829a2bc107f1def84055d9db2R9

Right, it only works with modules.

> Didn't we decide that we don't want to serve from /components/ anymore?

Yeah, but we still print it out when you run serve. It's still needed for 2.x use-cases, which we still support, so maybe we need to do something smart to detect what URL to suggest people visit. I think we should make sure the /components/ directory works for now, though.

packages/cli/templates/element/polymer-3.x/demo/index.html
```diff
@@ -8,8 +8,11 @@
 
     <script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 
-    <script type="module" src="../node_modules/@polymer/iron-demo-helpers/demo-pages-shared-styles.js"></script>
-    <script type="module" src="../node_modules/@polymer/iron-demo-helpers/demo-snippet.js"></script>
+    <script type="module">
+      import '@polymer/iron-demo-helpers/demo-pages-shared-styles';
```
bicknellr: `.js`

aomarks: The .js isn't actually needed, node resolution includes adding a .js. Have we typically been including them anyway?

bicknellr: I think so (modulizer doesn't remove them, fwiw) but it doesn't seem like we're totally consistent since some of our projects make use of node's `main` field: https://github.com/Polymer/pwa-starter-kit/blob/8da3dcc42472e99fa22d0a996362828d31dc70bb/src/components/my-app.js#L11-L13

bicknellr: IMO, we shouldn't have allowed these kinds of special behaviors from Node.js' specifier resolution in polyserve's rewriting and really hope this gets dropped in spec land.

PR #367, opened 2018-05-11
Correct the names of the new uncompiled presets

ranb2002:
Correct the names of the new uncompiled presets as build names must be unique


aomarks:
Good catch, thanks.

PR #370, opened 2018-05-11
polymer-cli 1.7.2 release cascade

usergenic:
- [ ] polymer-analyzer@3.0.1
  * Better support for handling errors early on in the analysis process (load,
    parse, scan). This should solve the vast majority of Internal Error warnings.
- [ ] polymer-project-config@4.0.1
  * Fix names of `uncompiled-bundled` and `uncompiled-unbundled` build presets.
- [ ] polymer-bundler@4.0.1
  * Added missing `babel-types` to dependencies instead of relying on it transitively.
- [ ] polymer-cli@1.7.2
  * Fix bug in `polymer-3-element` init template where `polymer/dom-module.js`
    could be loaded twice when serving from the polyserve `/components/`
    directory.
  * Fix names of `uncompiled-bundled` and `uncompiled-unbundled` build presets.


justinfagnani:
Agreed with @rictic here, let's try to keep the lower bounds as low as possible, unless we require a new feature or bugfix.

usergenic:
Force pushed a new PR commit, to prepare for release of just the four packages that describe unreleased changes in their CHANGELOG files.

PR #376, opened 2018-05-14
[wct] Moved server files to src, improved TypeScript configuration.

### LINKED ISSUE #221: [wct] Improving TypeScript configuration
Current TypeScript implementation has a few flaws:
- In the end compiled files are imported by a [js file](https://github.com/Polymer/tools/blob/master/packages/web-component-tester/runner.js), which strips all type definitions when WCT is used a module
- It does not follow the conventions which makes it harder for people to contribute (custom things always require learning something)

I'd like to:
- move files a little to follow `src`/`lib` convention
- move `runner.js` to `src` and change it to TypeScript
- change `main` field in `package.json`
- tweak `files` in `package.json` to make sure we publish only the required things

Is that ok with you guys?

merlinnot:
Changes in this PR:
- moved files from `runner` to `src`
- simplified root `tsconfig.json`
- set `tsc`'s `outDir` to `lib`
- changed `files` in `package.json` to include `lib`
- changed `main` field in `package.json`
- sorted & cleaned `files` in `package.json`
- rewritten main file in TypeScript (it was a `.js` file importing built files which was making it much harder to consume this package by a TypeScript program)
- removed `.npmignore` since NPM uses `.gitignore` by default anyways.
- improved `scripts` in `package.json` to simplify testing (compiling on the fly using `ts-node`)
- removed unnecessary Gulp tasks
- simplified `.gitignore` to reflect changes to the compilation and testing process
- removed unused `mocha.opts` (the correct one is still in `/test`)
- moved `prepublishOnly` script from `gulpfile.js` to `package.json`
- added `npm-run-all` as a dev dependency to simplify writing NPM scripts and make it compatible with both NPM and yarn
- moved `custom_typings` to `@types` to follow a convention & new TypeScript capabilities

Resolves #221


TimvdLippe:
This looks like a breaking change to me, so if we decide to merge this (I am generally in favor of this change), we have to decide what other breaking changes we want to ship and design the new major version of WCT accordingly.

merlinnot:
Why does it look like a breaking change to you? Do you think that people were importing some of the files directly? That's a bad practice in general, do we support it?

That's also why I didn't add a changelog entry, since it's not a user-facing change of the functionality.

merlinnot:
I've also added a detailed list of changes so it's easier to wrap your head around it :wink:

merlinnot:
Hi, can we do anything about this PR? WCT is next to unusable as a module when imported from TypeScript.

merlinnot:
Closing due to inactivity.

PR #380, opened 2018-05-14
[build] Pin babel-plugin-minify-guarded-expressions to 0.4.1

usergenic:
Latest version of this package `babel-plugin-minify-guarded-expressions` which is a dependency of `babel-preset-minify@0.4.0-alpha.caaefb4c` we have pinned will throw errors due to trying to get a parentPath from a path's parentPath in some cases where one does not exist and kills the entire optimize-streams.ts transform.

### DIFF
--- a/packages/analyzer/package-lock.json
+++ b/packages/analyzer/package-lock.json
@@ -208,9 +208,9 @@
 			"dev": true
 		},
 		"@types/node": {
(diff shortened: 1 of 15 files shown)


usergenic:
One test timeout unique to one vm, considering transient and merging PR.

PR #388, opened 2018-05-15
Adding workbox

prateekbh:
- Shifting PR from earlier repo to here


rictic:
Awesome! This will be a breaking change of the CLI, but it'll be really nice to get out there.

I wonder if we want to merge this into a branch? WDYT Justin?

justinfagnani:
Thanks for moving the PR here @prateekbh and the review @jeffposnick !

I think what we'll want for backwards compatibility is to keep `--add-service-worker` with no value to mean use sw-precache and read `sw-precache-config.js`, and make `--add-service-worker=workbox` do the new behavior here.

@prateekbh we could move this into a branch here and take over that part, or you could do it. Mostly I think it would involve making two separate service-worker worker modules and choosing one based on the value of the flag.

packages/build/custom_typings/workbox.d.ts
```diff
@@ -0,0 +1,69 @@
+interface ManifestEntry {
+  url: string;
+  revision: string;
+}
+
+declare module "workbox-build" {
```
justinfagnani: Since you all own Workbox, would it be better to move the typings over there?

packages/build/src/service-worker.ts
```diff
@@ -12,16 +12,16 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-/// <reference path="../custom_typings/sw-precache.d.ts" />
+/// <reference path="../custom_typings/workbox.d.ts" />
```
justinfagnani: I don't think this /// reference should be necessary. Try removing it?

packages/build/src/service-worker.ts
```diff
@@ -41,16 +41,6 @@ export interface AddServiceWorkerOptions {
  */
 // tslint:disable-next-line: no-any Turned off for user input.
 function fixDeprecatedOptions(options: any): AddServiceWorkerOptions {
```
justinfagnani: This function can be removed now it seems.

packages/build/src/service-worker.ts
```diff
@@ -100,78 +90,50 @@ export async function generateServiceWorkerConfig(
   options = Object.assign({}, options);
   const project = options.project;
   const buildRoot = options.buildRoot;
-  const swPrecacheConfig: SWConfig =
-      Object.assign({}, options.swPrecacheConfig);
+  const workboxConfig: WorkboxConfig =
+      Object.assign({}, {
```
justinfagnani: Use object spread instead of assign

packages/build/src/service-worker.ts
```diff
@@ -100,78 +90,50 @@ export async function generateServiceWorkerConfig(
...
-      replacePrefix = replacePrefix + '/' as PosixPath;
-    }
-    if (swPrecacheConfig.replacePrefix) {
-      console.info(
-          `Replacing service worker configuration's ` +
-          `replacePrefix option (${swPrecacheConfig.replacePrefix}) ` +
-          `with the build configuration's basePath (${replacePrefix}).`);
-    }
-    swPrecacheConfig.replacePrefix = replacePrefix;
+    workboxConfig.modifyUrlPrefix = Object.assign({}, workboxConfig.modifyUrlPrefix, {
```
justinfagnani: Use object spread

packages/build/src/test/service-worker_test.ts
```diff
@@ -174,56 +161,19 @@ suite('service-worker', () => {
                 assert.include(fileContents, '"bower_components/dep.html"');
                 assert.notInclude(fileContents, '"source-dir/my-app.html"');
               });
-        });
-
-    test(
-        'should add bundled precached assets when options.bundled is provided',
```
justinfagnani: We still want to make sure that we precache bundles, yes?

Dabolus:
Maybe I'm asking too much, but it would also be great to be able to use [injectManifest](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build.html#.injectManifest) instead of (or as an optional alternative to) [generateSW](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build.html#.generateSW) to make it possible to add some additional functionality to the SW instead of using it just for caching purposes, WDYT?

justinfagnani:
@prateekbh a deprecation warning would be great.

packages/build/custom_typings/workbox.d.ts
prateekbh: @jeffposnick Thoughts?

mercmobily:
@prateekbh This is awesome work -- I am starting a major porting to PWA (from the old PSK) and am making sure I am using the "latest everything" for it -- and workbox falls in that category!

prateekbh:
@justinfagnani I am fixing the tests and lets take it to a different branch. Once all looks good then wither you guys can add the flag or i'll create new files to help with the new workbox option.

Can you create a new branch and edit the PR please?

packages/build/custom_typings/workbox.d.ts
prateekbh: @jeffposnick I guess we can go ahead with keeping it here and update if we ever move this inside workbox

packages/build/src/test/service-worker_test.ts
prateekbh: Added back

packages/build/src/service-worker.ts
jsilvermist: I believe something like this?
```js
const workboxConfig: WorkboxConfig = {
  globDirectory: buildRoot,
  navigateFallback: path.relative(project.config.root, project.config.entrypoint),
  ...options.workboxConfig
};
```

packages/build/custom_typings/workbox.d.ts
jeffposnick: Looping back to this: the Workbox team's stance on `.d.ts` is reflected at https://github.com/GoogleChrome/workbox/pull/1436#issuecomment-396678604

None of the core maintainers are familiar with TypeScript today, and we don't feel comfortable with the guarantees of support that we'd signal if we hosted the definitions.

Either checking things into this repo, or alternatively hosting them on https://definitelytyped.org/, seems like the best course of action at this time.

prateekbh:
@justinfagnani @rictic:
I've modified the PR to use a different flag `addWorkboxServiceWorker` and shifted the changes to new files, so that it no more requires a major version bump. 😃

packages/build/src/polymer-build.ts
```diff
@@ -11,3 +12,5 @@ export {getOptimizeStreams, OptimizeOptions} from './optimize-streams';
 export {PolymerProject} from './polymer-project';
 export {addServiceWorker, generateServiceWorker} from './service-worker';
 export {AddServiceWorkerOptions} from './service-worker';
+export {addWorkboxServiceWorker, generateWorkboxServiceWorker} from './service-worker-workbox';
+export {AddServiceWorkerWorkboxOptions} from './service-worker-workbox';
```
moderndeveloperllc: @prateekbh Small nitpick. There is `addWorkboxServiceWorker` then `AddServiceWorkerWorkboxOptions`. Can you standardize to either having "Workbox" before or after "ServiceWorker"? It reads better if the placement of the additional word is consistent.

prateekbh: Thanks. Will do this by EOD

moderndeveloperllc: @prateekbh Ping.

prateekbh: Sry! got caught up! doing it now

prateekbh: @moderndeveloperllc done

moderndeveloperllc:
@justinfagnani Can y'all chime in on this and see if the PR now makes sense? Wondering if perhaps @prateekbh needs to squash the commits?

JosefJezek:
We are waiting for Workbox... Please merge asap. :hammer_and_wrench:

prateekbh:
@justinfagnani @rictic i reloved the last comment about removing `object.assign`.
I am not sure t=what is causing the tests to fail, can you some help

packages/build/src/service-worker-workbox.ts
```diff
@@ -0,0 +1,173 @@
...
+      getBundledPrecachedAssets(project) :
+      getPrecachedAssets(depsIndex, project);
+
+  if (workboxConfig.globPatterns === undefined) {
+    staticFileGlobs = staticFileGlobs.concat(precachedAssets);
+    staticFileGlobs = staticFileGlobs.map((filePath: string) => {
+      if (filePath.startsWith(project.config.root)) {
+        filePath = filePath.substring(project.config.root.length);
+      }
+      return removeLeadingSlash(filePath);
```
jeffposnick: I'm taking a look at the Windows CI test failure at https://ci.appveyor.com/project/Polymer/tools/build/1.0.1475#L1409 and one thing that stands out is that the string `'\\shell.html'` is being used as a glob pattern. That shouldn't be what's used, since glob patterns always use `/` as a path separator, even on windows.

I feel like the `staticFileGlobs.concat(precachedAssets)` line a few lines up might be introducing paths with `'\\'` characters into the list of glob patterns, and then this line of code isn't sufficient to remove that leading `'\\'`.

But in general, all glob patterns need to use `'/'`, regardless of where the separator occurs.

prateekbh:
@rictic @justinfagnani can you please restart task: https://travis-ci.org/Polymer/tools/jobs/414764209

its a connection timeout error. All other tests are passing

packages/build/src/service-worker-workbox.ts
prateekbh: Thanks for the suggestion @jeffposnick , fixed it

prateekbh:
So in theory all tests are paying bit every other run of Travis fails on some or other connection timeout.

Waiting from someone in polymer cli team to look into it

abdonrd:
Any news about merging this PR? Thanks!

JosefJezek:
Any news?

mercmobily:
Has this PR been abandoned? I was looking forward to it!

prateekbh:
Well I never heard back from the polymer cli team. I am still open to get this in.

mercmobily:
@justinfagnani Would you be willing to merge this after review?

HighOnDrive:
The world is waiting with bated breath...

mercmobily:
@prateekbh Still willing to work on it? Can you think of a reason why this is not  moving forward?

vladimirbrasil:
Guys, web pages are becoming so cool that we might just use it "to the bones". No tools. lit-element might be the only remaining. Don't know. Service workers are becoming so cool that we must use it directly, without "doormans" or intermediaries. It was a time, comprehensive time, where "service workers" needed to be "abstracted away" and keep distant and handled by others. No, time's changed. This request is a photograph of this old time: people begging to use workbox, to get closer to their own service worker; while programmer gurus don't have time to do this big change of behavior. It is indeed a big change of behavior, isn't it? It is still happening. Service worker craftsmanship must be directly handled by ourselves, lovely programmers. Let's not keep service workers distant anymore. Let's deal with cache changing without giving away, not anymore, our baby, our beloved service worker. It is incredibly doing each time better stuff. Why use a server?, we can start to ask more times, with beautiful processors and service workers each day more powerful.

Ignore the poetry and the sentiment. I am just saying that service workers are now babies that we can not afford to handle to others, to keep it distant, even with good reasons like dealing with new builds and cache invalidation. Not an expert, you should already have noticed. Can only give this "general" opinion and sentiment.

prateekbh:
The PR is extremely old and outdated closing this one

mercmobily:
@prateekbh  Wait... really?
Am I right in saying that this is a fully functional PR, and all tests should pass (apart from technical hiccups), it was even checked by developers, and yet it's being dropped?

@justinfagnani  @rictic  are there other plans to move to Workbox?

prateekbh:
well yes most tests passed, but this is over an year old PR. workbox has rolled out many versions after this and i have no idea how much work would go in to actually get it updated and in.

I would happily do a fresh one if there's buy in from polymer team

Issue #229, opened 2018-04-25
[wct] Memory leak caused by iframes


merlinnot:
All iframes are stored in a memory when tests are running. This is not necessary, we can destroy an iframe after collecting tests results for a suite.

With a high number of tests it results in arbitrary large memory consumption. Also reported agains the old repository:
https://github.com/Polymer/web-component-tester/issues/579
https://github.com/Polymer/web-component-tester/issues/545

kevinpschaaf:
Cross posting from https://github.com/Polymer/web-component-tester/issues/579#issuecomment-405386382:

It appears the leak was unintentional.  Although the top runner keeps a list of `ChildRunner` objects resident in memory, the `ChildRunner` nulls out its iframe after tests are complete -- so the intention was that the iframe is not retained.

However there is a (not well understood by me) "`share`" field on the child runner which was not being nulled at the completion of the runner, which was retaining the iframe.  Clearing this at the same time the iframe field is cleared results in memory correctly being freed and dramatically lower overall memory usage.

justinfagnani:
@kevinpschaaf does the "share" object have a reference back into the iframe?

kevinpschaaf:
@justinfagnani Yes.  I don't fully understand the reference path, but via the memory profiler we traced the reference set here back to what was causing the iframes to be retained: https://github.com/Polymer/tools/blob/master/packages/web-component-tester/browser.js#L629

justinfagnani:
That's what confused me. That line shouldn't cause a reference to be held, unless `share` had a reference to the iframe. If `this.iframe` is null, that line will error. So either `share` has to have a reference to the iframe, or there some other reference to it or it's contentWindow or some other DOM node.

justinfagnani:
Oh, and unfortunately `share` is completely untyped, so we can't see what's going on from source.

kevinpschaaf:
@justinfagnani So, it turns out a completely empty object created one iframe but held in another iframe is enough to keep the iframe it was created in from being gc'ed.

This is exactly what happens with the `share` property: The `share` property of the inner WCT is [initiated with an object literal `{}`](https://github.com/Polymer/tools/blob/master/packages/web-component-tester/browser.js#L1789), and a reference of that is set on the `ChildRunner` instance's `share`  in property in the outer iframe.  That alone is enough to prevent the iframe from being gc'ed.  Clearing the `share` property on the `ChildRunner` when the iframe is cleaned up fixes it.

PR #393, opened 2018-05-16
[wct] Decrease possibility of child suites (iframes) being retained in memory

merlinnot:
Event listeners were not removed after an execution of a suite.

This is a step towards resolving #229.


packages/web-component-tester/browser/childrunner.ts
```diff
@@ -10,6 +10,9 @@
  */
 import * as util from './util.js';
 
+const DOM_CONTENT_LOADED_EVENT_NAME = 'DOMContentLoaded';
+const IFRAME_ERROR_EVENT_NAME = 'error';
```
rictic: Local style is `lowerCamelCase` for constants.

packages/web-component-tester/browser/childrunner.ts
```diff
@@ -167,25 +194,34 @@ export default class ChildRunner {
...
-    setTimeout(function() {
-      this.iframe.parentNode.removeChild(this.iframe);
-      this.iframe = null;
-    }.bind(this), 1);
+    if (this.iframe) {
+      // Be safe and avoid potential browser crashes when logic attempts to
+      // interact with the removed iframe.
+      setTimeout(() => {
+        const { iframe } = this;
+        iframe.removeEventListener(
```
rictic: For this sort of cleanup work, I prefer an array of callbacks to call as a private member on the class. That way the cleanup work can be registered at the same time and in the same place as the stuff that needs cleanup.

That would also let us avoid the repetitive file-level constants for event names

merlinnot:
@rictic I've changed the approach a little, I hope you'll find it aligned with your concept.

merlinnot:
@rictic a friendly ping :)

merlinnot:
@rictic Travis throws `session not created exception: Chrome version must be >= 66.0.3359.0`.

usergenic:
Thanks for PR @merlinnot - Integrating now.

PR #399, opened 2018-05-17
Hoist typescript and tslint to top-level package.json

justinfagnani:


justinfagnani:
Updated the package-locks with `npm nuke` which shaved 600MB off the total install size after hoisting.

PR #424, opened 2018-05-31
Fix: WCT CLI arg standardizing

### LINKED ISSUE #416: polymer test `--config-file` is ignored
`polymer test --help` says:
~~~
  --config-file string             Config file that needs to be used by wct. ie: wct.config-sauce.js             
~~~~

Looking through https://github.com/Polymer/tools/blob/219ab4f3f9f8773e75f8c6181109e8966082b9af/packages/cli/src/commands/test.ts#L139-L157: The argument is passed directly to wct's cli.

That one however expects `configFile`: https://github.com/Polymer/tools/blob/219ab4f3f9f8773e75f8c6181109e8966082b9af/packages/web-component-tester/runner/config.ts#L349-L352

I'm not sure how to resolve this, as `polymer-cli` has both options with camelCase and with dash-case.

Right now though the result of this is that `--config-file` doesn't work, and I need to do something like this in Travis: `mv travis-wct.conf.json wct.conf.json`. Clearly ugly :/

stramel:
Supersedes https://github.com/Polymer/web-component-tester/pull/698

Fixes #416 https://github.com/Polymer/polymer-cli/issues/919 https://github.com/Polymer/web-component-tester/issues/652 

@usergenic

### DIFF
--- a/packages/web-component-tester/CHANGELOG.md
+++ b/packages/web-component-tester/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 * Updated Mocha to v5
+* Fixed incorrect CLI arguments
 
 ## 6.6.0 - 2018-05-08
 * Dropped support for node v6. This is a soft break, as we aren't
--- a/packages/web-component-tester/README.md
+++ b/packages/web-component-tester/README.md
@@ -184,8 +184,8 @@ you based on the suites you ask it to load._
 
 The `wct` command line tool will pick up custom configuration from a
 `wct.conf.json` file located in the root of your project.
-Or, you can specify your own file via the `--configFile` command line option.
-Example: `--configFile my.wct.conf.js`
+Or, you can specify your own file via the `--config-file` command line option.
+Example: `--config-file my.wct.conf.js`
 If you define your own configuration file, make sure you also provide the correct root if needed.
 By default it will use the directory in which the configuration file is found as rootpath, which can result in errors if the file is in a sub directory.
 
--- a/packages/web-component-tester/runner/config.ts
+++ b/packages/web-component-tester/runner/config.ts
@@ -339,6 +339,7 @@ const ARG_CONFIG = {
   },
   simpleOutput: {
     help: 'Avoid fancy terminal output.',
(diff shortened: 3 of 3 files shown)


stramel:
Or is it a fix since it should have been this way in the first place? 🤔 😛

merlinnot:
😂

usergenic:
hmm https://github.com/search?q=polymer+test+--configFile&type=Code

stramel:
@usergenic I wonder if I could set `alias: 'configFile'`?

usergenic:
@stramel that occurred to me too. haven't tested it but thats probably the way to go.

stramel:
Yeah, I won't be able to test until I get home. But if it works I will update the PR and add a note

merlinnot:
Maybe a deprecation warning on top of this? So we can actually drop it some day?

stramel:
:/ alias can only be a single character... I will look into a different way to resolve that issue.

stramel:
@usergenic Perhaps this is just a happy coincidence but this isn't a breaking change oddly enough. It is "breaking" in the sense that we are deprecating the `configFile` and `simpleOutput` option. However, `nomnom` the arg parser that is used in `web-component-tester` will match `configFile` and `config-file` after these changes. Where as previously it did not match `config-file` as `configFile`.

Technically speaking any arg listed camelCase in the definition list will work camelCase even with `full` specified as dash-case. 

**NOTE:** `nomnom` is deprecated and would probably be useful to move to `command-line-args` like the other packages on the next major breaking change release and coordinate that with the remove of camelCase arg support.

usergenic:
@stramel whoa!  how about that!  okay, that's great work looking into that.  After we merge this I'll add a test or something to ensure handling of both forms in the event someone decides to upgrade wct's arg parsing without knowing all this.

usergenic:
I'll get this merged before tomorrow as soon as I can set aside a moment to craft that test.

usergenic:
Had some other issues with CLI release branch on Friday; Will get this merged *today* (if possible!)

stramel:
No worries! Thanks again @usergenic!

usergenic:
@stramel this is getting released shortly after https://github.com/Polymer/tools/pull/515 gets approved. :)

PR #435, opened 2018-06-05
Import dom5

usergenic:
Import the dom5 package into the monorepo

Opportunistically:
- Pins typescript to ~2.8.3 because of issue with 2.9.x reading the parallel index.d.ts file in parse5 and we don't want that right now, thank you, Typescript.
- Adds missing clone package to editor-service devDependencies which was in use in one test file


usergenic:
FYI - had to updated goldens to CLI integration tests because of update from babel 7.0.0-beta.47 to 7.0.0-beta.49 resulted in slightly different compiled output.

usergenic:
appveyor failure is due to Polymer/shop lit-element branch having a git repo dependency in its package.json -- Merging this PR in spite of the integration test failure, which needs to be fixed in Polymer/shop

PR #436, opened 2018-06-05
[all] Hoist mocha

justinfagnani:


packages/web-component-tester/wct-browser-legacy/package.json
```diff
@@ -33,7 +33,6 @@
     "async": "^1.5.2",
     "chai": "^3.5.0",
     "lodash": "^3.10.1",
-    "mocha": "^3.4.2",
```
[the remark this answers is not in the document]
justinfagnani: fixed

packages/web-component-tester/wct-browser-legacy/package-lock.json
```diff
@@ -57,25 +57,6 @@
 			"resolved": "https://registry.npmjs.org/async/-/async-1.5.2.tgz",
```
rictic: Update the wct-browser-legacy package-lock? Or, uh, un-update it?

justinfagnani: yeah...

justinfagnani:
@rictic PTAL, and let's hope tests go green

rictic:
There's failures in bundler. These look like some changes I added that I wasn't certain about, where we needed to be more careful about the types of documents we had in a number of places.

It's possible that these tests were always failing, and the failures were just being swallowed previously.

usergenic:
@rictic the failures are fixed in master. Vscode-uri had changed something that broke the way bundler was extending/generating new urla

justinfagnani:
@rictic @usergenic just rebased on master to pick up those fixes

PR #439, opened 2018-06-06
Remove unused callback argument to wct.cli.run and wct.cli.startSauceTunnel

### LINKED ISSUE #438: wct-st exits upon sauce tunnel creation

azakus:
Also remove promise in wct-st, as the promise resolves immediately on
sauce tunnel creation, killing the new sauce tunnel

Fixes #438

### DIFF
--- a/packages/web-component-tester/bin/wct
+++ b/packages/web-component-tester/bin/wct
@@ -14,9 +14,7 @@ process.title = 'wct';
 
 resolve('web-component-tester', {basedir: process.cwd()}, function(error, path) {
   var wct = path ? require(path) : require('..');
-  var promise = wct.cli.run(process.env, process.argv.slice(2), process.stdout, function (error) {
-    process.exit(error ? 1 : 0);
-  });
+  var promise = wct.cli.run(process.env, process.argv.slice(2), process.stdout);
   if (promise) {
     promise.then(() => process.exit(0), () => process.exit(1));
   }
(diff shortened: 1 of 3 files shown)


packages/web-component-tester/bin/wct
```diff
@@ -14,9 +14,7 @@ process.title = 'wct';
 
 resolve('web-component-tester', {basedir: process.cwd()}, function(error, path) {
   var wct = path ? require(path) : require('..');
-  var promise = wct.cli.run(process.env, process.argv.slice(2), process.stdout, function (error) {
-    process.exit(error ? 1 : 0);
-  });
+  var promise = wct.cli.run(process.env, process.argv.slice(2), process.stdout);
```
justinfagnani: on the next line... do you know why this call might not return a promise?

justinfagnani:
I think this change constrains the `wct` command to only working with more recent versions of wct that have the Promise-based API. That's ok, but we should possibly check the version of wct that we resolve to and give a good error message if it's too old.

Or we could just rip out this behavior because it's been fairly confusing in the past.

packages/web-component-tester/bin/wct
azakus: Not a clue.

usergenic:
Expect error in polymer element 3.x test of polymer-cli to be transient.  Merging PR.

PR #441, opened 2018-06-06
Run esm-amd-loader on Sauce when configured.

aomarks:
Also replace SKIP_REMOTE_BROWSERS env var with WCT_SAUCE.

### DIFF
--- a/packages/esm-amd-loader/test/package.json
+++ b/packages/esm-amd-loader/test/package.json
@@ -7,7 +7,7 @@
     "clean": "rm -rf lib/",
     "lint": "tslint --project . --format stylish",
     "build": "npm run clean && tsc",
-    "test:wct": "npm run lint && npm run build && wct"
+    "test:wct": "npm run lint && npm run build && bash run-wct.sh"
   },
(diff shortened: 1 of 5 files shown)


packages/esm-amd-loader/test/package.json
```diff
@@ -7,7 +7,7 @@
     "clean": "rm -rf lib/",
     "lint": "tslint --project . --format stylish",
     "build": "npm run clean && tsc",
-    "test:wct": "npm run lint && npm run build && wct"
+    "test:wct": "npm run lint && npm run build && bash run-wct.sh"
```
usergenic: How does test:wct get invoked?  Shouldn't this be test:integration?

aomarks: esm-amd-loader's `test:integration` command cds to this package and calls this script. i thought that made more sense, since it's the integration test for that package, which just happens to be implemented in a separate package.

usergenic: Got it.  Thanks!

PR #446, opened 2018-06-07
Updated url-utils to appent the path by extension without using private APIs that have since changed.

usergenic:
This fix was part of https://github.com/Polymer/tools/pull/442 but that PR is stalled due to a problem with wct-local not shutting down selenium servers and the whole thing, well... stalling when integration tests run.  Need to fix that before PR can merge.


packages/bundler/src/url-utils.ts
```diff
@@ -27,10 +27,11 @@ import constants from './constants';
  *     'file:///something/something.html_omg.js?ponies'
  */
 export function appendUrlPath(url_: string, extension: string): string {
-  const uri = Uri.parse(url_);
-  uri['_path'] = `${uri.path}${extension}`;
-  return uri.toString();
+  const {scheme, authority, path, query, fragment} = Uri.parse(url_);
```
justinfagnani: Is there a reason to list all the individual fields? How do we know we got all the important ones? I'd feel more confident with just updating the path...

usergenic: https://github.com/Microsoft/vscode-uri/blob/master/src/index.ts#L684

The reason to list them is that technically these are the only fields that should be in the input to `Uri.from()` whereas the Uri instance returned from a parse call contains more stuff according to typings and then we get compile errors.

usergenic: Can get around it with Object.assign but that's only because its a typings cheat.

usergenic: And object spread syntax returns error about unsupported fields like fsPath

justinfagnani: ok, makes sense if the types are different.

usergenic:
Due to travis being delayed, the other PR actually finally got past its CI and merged before this one did.  So just closing this out.

PR #448, opened 2018-06-08
[all] Hoist chai

justinfagnani:
Depends on #436


packages/editor-service/package.json
```diff
@@ -21,11 +21,13 @@
     "format": "find src test | grep \"\\.js$\\|\\.ts$\" | xargs ./node_modules/.bin/clang-format --style=file -i"
   },
   "devDependencies": {
+<<<<<<< HEAD
```
rictic: merge failure :X

justinfagnani:
@rictic ptal

package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --ci",
     "build": "lerna run build --stream",
-    "nuke": "rm -rf package-lock.json node_modules && npm install && lerna exec 'rm -f package-lock.json npm-shrinkwrap.json' && lerna clean --yes && lerna bootstrap",
+    "nuke": "rm -rf package-lock.json node_modules && npm install && lerna exec 'rm -f package-lock.json npm-shrinkwrap.json' && lerna clean --yes && lerna exec --stream -- npm install && lerna bootstrap",
```
TimvdLippe: @usergenic Thanks for the tip! I actually had to update the script, as `npm run bootstrap` generates incorrect lock files. That's why I had to make all these PRs to make `audit` run. I have updated the script to first run `npm install` (which generates the correct lockfile) and then it runs `bootstrap`, such that the symlinks are in place.

justinfagnani:
Thanks for the updates @TimvdLippe !

PR #511, opened 2018-06-12
Make esm-amd-loader dynamic require 404 test not flaky.

### LINKED ISSUE #507: esm-amd-loader 404s test is flaky
The "calls error callback only once on multiple 404s" test appears to be flaky. I keep seeing different browsers fail this test on CI:

```
OS X 10.13 safari latest ✖ static/y/suite.html » dynamic require » calls error callback only once on multiple 404s
  expected 0 to equal 1
    <unknown> at                      AssertionError at /components/wct-browser-legacy/node_modules/chai/chai.js:5260:0
    <unknown> at                              assert at /components/wct-browser-legacy/node_modules/chai/chai.js:259:0
    <unknown> at                               equal at /components/wct-browser-legacy/node_modules/chai/chai.js:2006:0
    <unknown> at   http://localhost:8000/components/ at polymer/esm-amd-loader-test/lib/suite.js:225:0
chrome 67                Tests passed
Windows 8.1 IE 11        Beginning tests via http://localhost:8000/components/@polymer/esm-amd-loader-test/generated-index.html?cli_browser_id=0
firefox 60               Tests passed
OS X 10.13 safari latest Tests failed: 1 failed tests
Windows 10 microsoftedge latest Tests passed
Windows 8.1 IE 11        Tests passed
Test run ended in failure: 1 failed tests
```

aomarks:
Previously this test used a 1 second timeout to wait to see how many times the error callback was called, which was flaky on CI. Now we capture 404 errors and check callback calls after two 404s.

Fixes #507

(Note first commit is just clang formatting).

### DIFF
--- a/packages/esm-amd-loader/test/src/suite.ts
+++ b/packages/esm-amd-loader/test/src/suite.ts
@@ -268,19 +268,25 @@ suite('dynamic require', () => {
   });
 
   test('calls error callback only once on multiple 404s', (done) => {
-    let numErrors = 0;
+    let num404s = 0;
+    let numCallbackCalls = 0;
+
+    window.addEventListener('error', on404, true);
+
+    function on404() {
+      if (++num404s === 2) {
+        window.removeEventListener('error', on404);
+        assert.equal(numCallbackCalls, 1);
+        done();
+      }
+    }
 
     define(['require'], (require: any) => {
       require(
           ['./not-found-a.js', './not-found-b.js'],
           () => assert.fail(),
-          () => numErrors++);
+          () => numCallbackCalls++);
     });
-
-    setTimeout(() => {
-      assert.equal(numErrors, 1);
-      done();
-    }, 1000);
   });
 });
 
--- a/packages/esm-amd-loader/test/src/suite.ts
+++ b/packages/esm-amd-loader/test/src/suite.ts
@@ -407,16 +407,16 @@ suite('html imports', () => {
   }
 
   function testImport(href: string, expectedOrder: string[], done: () => void) {
-     // Each time an amd module in the chain is executed, it registers itself.
-     // If we've reached the length of modules we are expecing to be loaded,
-     // we check if the right modules were loaded in the expected order
-     window.addExecutedForImport = (name: string) => {
-       window.executionOrder.push(name);
-       if (window.executionOrder.length === expectedOrder.length) {
-         assert.deepEqual(window.executionOrder, expectedOrder);
-         done();
-       }
-     };
+    // Each time an amd module in the chain is executed, it registers itself.
+    // If we've reached the length of modules we are expecing to be loaded,
+    // we check if the right modules were loaded in the expected order
+    window.addExecutedForImport = (name: string) => {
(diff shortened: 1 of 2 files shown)


packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -268,19 +268,25 @@ suite('dynamic require', () => {
...
 
   test('calls error callback only once on multiple 404s', (done) => {
-    let numErrors = 0;
+    let num404s = 0;
+    let numCallbackCalls = 0;
+
+    window.addEventListener('error', on404, true);
+
+    function on404() {
+      if (++num404s === 2) {
```
rictic: Prefer inert conditionals.

```typescript
num404s++;
if (num404s === 2) {
  ...
```

packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -268,19 +268,25 @@ suite('dynamic require', () => {
   });
 
   test('calls error callback only once on multiple 404s', (done) => {
-    let numErrors = 0;
+    let num404s = 0;
+    let numCallbackCalls = 0;
+
+    window.addEventListener('error', on404, true);
```
rictic: Hm, I don't think this is the expected behavior. `require` should not throw a global error, it should only call the failure callback shouldn't it?

packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -268,19 +268,25 @@ suite('dynamic require', () => {
...
+        done();
+      }
+    }
 
     define(['require'], (require: any) => {
       require(
           ['./not-found-a.js', './not-found-b.js'],
           () => assert.fail(),
-          () => numErrors++);
+          () => numCallbackCalls++);
```
rictic: Hm, what we really want to do is wait for the first error callback, and consider the test a success if the error callback isn't called again within the next little while (i.e. using setTimeout).

That should be non-flaky.

packages/esm-amd-loader/test/src/suite.ts
rictic: Oh, is this the browser doing this? Emitting a window error?

aomarks: Yeah, you can capture the browser network errors by adding the capture phase listener (the true).

packages/esm-amd-loader/test/src/suite.ts
aomarks: IIUC that would be more error prone (since network latency between the two fetches could cause the test to incorrectly pass) and slower (since we have to wait for the timeout).

packages/esm-amd-loader/test/src/suite.ts
aomarks: Done.

packages/esm-amd-loader/test/src/suite.ts
aomarks: I did need to add a tick to let the error handlers get called though. Seems reliably green on sauce now, will see what CI says.

PR #512, opened 2018-06-13
Fix push manifest generation bugs affecting bower_components and ES modules

### LINKED ISSUE #433: addPushManifest treats app dependencies as siblings
Shop#master (Polymer 2.0) uses the `es6-unbundled` preset which uses the `addPushManifest` build configuration. Today, running `polymer build` generates a push-manifest.json that references dependencies as siblings:

```js
{
  "src/shop-app.html": {
    "../polymer/polymer-element.html": {
      "type": "document",
      "weight": 1
    },
```

### Expected

Previously, they are correctly referenced as files within `bower_components`:

```json
{
  "src/shop-app.html": {
    "bower_components/polymer/polymer-element.html": {
      "type": "document",
      "weight": 1
    },
```

This regressed sometime during the run-up to the Polymer 3.0 release.

aomarks:
This PR fixes two issues with push manifest generation:

1) Projects with bower_components were getting paths relative to the package root instead of the project root. This seems to have been a regression in 1.7 (#433).

2) Projects with ES modules were crashing on push manifest generation because it was unable to resolve bundler's special sub-bundle URLs (e.g. `Unable to get document file:///private/var/folders/wz/fv09mcr569gdkbzqct263l_r003hs9/T/tmp-84970LqLvdcKS3nbH/index.html>external#1>mod1.js>es6-module: unknown`).

### DIFF
--- a/packages/build/CHANGELOG.md
+++ b/packages/build/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Fix incorrect relative paths to the component directory in push manifests.
+* Fix push manifest generation crash with ES module projects.
 <!-- Add new, unreleased changes here. -->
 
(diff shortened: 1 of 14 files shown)


usergenic:
Need to fix; `test:unit` failing reliably due to change.

PR #515, opened 2018-06-13
Prepare release of wct-sauce, polymer-build, web-component-tester and polymer-cli.

usergenic:
- [ ] wct-sauce@2.0.2
- [ ] web-component-tester@6.7.0
- [ ] polymer-build@3.0.2
- [ ] polymer-cli@1.7.4

Note that polymer-cli has a bug fix to problem in 1.7.3 polymer-cli release that appears to have been result of package update to command-line-args which is more strict on args without explicit `multiple: true`.


packages/cli/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 * Fix push manifest generation crash with ES module projects.
 <!-- Add new, unreleased items here. -->
 
+## v1.7.4 [06-13-2018]
```
aomarks: 06-18-2018

packages/web-component-tester/package.json
```diff
@@ -84,7 +84,7 @@
   "optionalDependencies": {
     "update-notifier": "^2.2.0",
     "wct-local": "^2.1.0",
-    "wct-sauce": "^2.0.0"
+    "wct-sauce": "^2.0.2"
```
justinfagnani: I think we should move away from using optionalDependencies.

PR #524, opened 2018-06-15
Configure module directory in bare module specifier rewrite

LarsDenBakker:
This PR passes the module directory to the node resolution module. It defaults to node_modules, but when other folders are used (ex. bower_components, or any other custom folder) it fails to find the right modules.


LarsDenBakker:
Found an easier way to do this: https://github.com/Polymer/tools/pull/526

Issue #521, opened 2018-06-15
Allow bare specifier imports from bower_components


LarsDenBakker:
We're using modules, but are still using bower for now. However, module in bower_components are not picked up when rewriting bare import specifiers. Is it possible to add this?

PR #526, opened 2018-06-16
Add bower_components to list of module directories

LarsDenBakker:
The resolve module only resolves modules from the node_modules directory. Adding bower_components to it let's us use modules with bower, and matches the behavior of tools like Typescript.

Fixes https://github.com/Polymer/tools/issues/521


daKmoR:
anyone got time to review this?

PR #527, opened 2018-06-17
Use native Node promisify, to aviod issues with Node 10

HitkoDev:
Node 8 comes with a builtin promisify function, which has proper type declarations and is well-tested. `promisify-node` breaks with Node 10, since we don't need support for Node < 8, we should use the builtin function. https://github.com/Polymer/tools/blob/ec5693d318fa4766293176409fa5df8790d8d111/package.json#L29


TimvdLippe:
Ref: #496

HitkoDev:
I'd definitely go with a native option, especially considering `promisify-node` is one of the least popular promisify solutions on NPM, and we have to rely on unmaintained type definitions to use it.

usergenic:
This is pretty great!  Thanks, @HitkoDev -- I'm currently working on Node 10-ifying the tools suite and I really appreciate this PR.

PR #543, opened 2018-06-26
Do not compile NPMPackages which are used during testing

### LINKED ISSUE #538: [polyserve] Serving extremely slow on sinon package bundle
Serving the sinon.js bundle with `polyserve` is extremely slow (>12 seconds).

To reproduce:

1. Check out branch `fix-audit-wct-browser-legacy`
1. Install the dependencies per usual instructions
1. Run `polymer serve` in the root of this repository
1. Open http://127.0.0.1:8081/components/polymer-tools/packages/wct-browser-legacy/node_modules/sinon/pkg/sinon.js (check your port)
1. Observe that loading this particular file is extremely slow

Even when running with `polymer serve --compile never`, it remains slow. Note that other files (for example http://127.0.0.1:8081/components/polymer-tools/packages/wct-browser-legacy/node_modules/sinon/History.md) are instantly loading. Also note that after you refresh the file, it is instant again.

This leads me to believe that there is a performance issue with reading/writing ASTs to files on first load.

This issue is blocking the migration of #533

TimvdLippe:
By moving this enumeration from WCT to Polyserve, we can now make sure we do not compile these dependencies (they do not need it anyway).

Fixes #538


packages/polyserve/src/compile-middleware.ts
```diff
@@ -38,6 +38,45 @@ const compileMimeTypes = [htmlMimeType, ...javaScriptMimeTypes];
...
+  /**
+   * JS entrypoints relative to packageName e.g. lodash/index.js would simply
+   * be ['index.js'] and myPackage/dist/addon.js and myPackage/lib/core.js would
+   * be ['dist/addon.js', 'lib/core.js']
+   */
+  jsEntrypoint: string;
+}
+
+// scripts to be injected into the running test
+export const ENVIRONMENT_SCRIPTS: NPMPackage[] = [
```
aomarks: What if we checked for a URL parameter like ?nocompile, and have WCT add that to the URLs it injects? Maybe that separates concerns better?

TimvdLippe: Oooh I like that solution. Will take a stab at that 👍

usergenic: +1 to not adding more WCT-specific content directly to polyserve.

TimvdLippe: Updated :+1: I will update the WCT stuff accordingly in the other PR to minimize noise.

packages/polyserve/src/compile-middleware.ts
```diff
@@ -72,6 +74,9 @@ export function babelCompile(
       if (isPolyfill.test(request.url)) {
         return false;
       }
+      if (shouldNotCompile.test(request.url)) {
```
aomarks: Maybe use `('nocompile' in request.query)` or similar so that `nocompile` doesn't have to be the first query parameter (e.g. `foo&nocompile`).

aomarks: https://expressjs.com/en/4x/api.html#req.query

TimvdLippe: Ah, TIL. Done :+1:

TimvdLippe:
@aomarks Right, I always forget that. Have to add that PR_TEMPLATE soonTM :wink:

PR #555, opened 2018-06-26
Add browser-capabilities entry for Chrome on iOS.

aomarks:
Treat Safari on iOS as though it were Safari Mobile. Also use iOS version, instead of Safari version, to better handle Safari Mobile.

### DIFF
--- a/packages/browser-capabilities/CHANGELOG.md
+++ b/packages/browser-capabilities/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
-<!-- ## Unreleased -->
+## Unreleased
+* Chrome for iOS is now handled correctly, by delegating to the corresponding
+  Safari feature matrix.
+* Safari for iOS is handled more accurately, using iOS version.
 <!-- Add new, unreleased changes here. -->
 
 ## [1.1.0] - 2018-05-03
--- a/packages/browser-capabilities/src/browser-capabilities.ts
+++ b/packages/browser-capabilities/src/browser-capabilities.ts
@@ -51,17 +51,6 @@ const browserPredicates: {
   'Chrome': chrome,
   'Chromium': chrome,
   'Chrome Headless': chrome,
-  'ChromeiOS': {
-    // TODO(aomarks) Capabilities for Chrome on iOS. We probably need to use the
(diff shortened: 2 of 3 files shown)


keanulee:
I'm thinking that for iOS, we should reply on mobile OS version exclusively. That way it should work across Chrome and Safari on iOS. After all, on iOS browser capabilities is determined by OS version. (Haven't thought about how the code would be structured, I'm in a meeting 😉 )

aomarks:
> I'm thinking that for iOS, we should reply on mobile OS version exclusively. That way it should work across Chrome and Safari on iOS. After all, on iOS browser capabilities is determined by OS version. (Haven't thought about how the code would be structured, I'm in a meeting 😉 )

We can easily match on OS version, I just need to know what the matrix looks like.

keanulee:
Based on [popular iOS UA strings](https://developers.whatismybrowser.com/useragents/explore/operating_system_name/ios/), all minor releases of the same major iOS version (since iOS 6) always maps to the same Safari version:

iOS version|Safari version
-|-
8_1|8.0
9_1|9.0
9_3|9.0
9_3_2|9.0
9_3_5|9.0
10_2_1|10.0
10_3_3|10.0
11_0|11.0
11_1_2|11.0
11_2_1|11.0
11_2_6|11.0
11_3|11.0

Given this, I'm assuming our [browser capabilities for Mobile Safari](https://github.com/Polymer/tools/blob/master/packages/browser-capabilities/src/browser-capabilities.ts#L77) is actually meant to match iOS version since you wouldn't be able to differentiate minor Safari versions - [caniuse for http/2 seems to reflect iOS 9.2 as well](https://caniuse.com/#feat=http2). So I think we should keep the same version numbers in `'Mobile Safari'`, but just match on iOS version instead (with the caveat being 11_3 may be the last version we will see moving forward due to frozen UA).

aomarks:
@keanulee How does this look?

aomarks:
> Code looks good. Suggest adding additional test cases for iOS 9 (e.g. that pwa-starter-kit bug with iOS 9/Chrome 62) and iOS 11_3 (supports SW/modules)

Done.

PR #557, opened 2018-06-28
[analyzer] Resolve `as` style module exports

aomarks:
Resolves what `foo` is in `export {foo as bar}`.

### DIFF
--- a/packages/analyzer/src/model/reference.ts
+++ b/packages/analyzer/src/model/reference.ts
@@ -122,6 +122,17 @@ function resolveScopedAt<K extends keyof FeatureKindMap>(
     const exportedIdentifier = getExportedIdentifier(path.node, identifier);
     return resolveThroughImport(path, exportedIdentifier, document, kind);
   }
+
+  // Handle cases like `export {Foo as Bar}`.
+  if (babel.isExportNamedDeclaration(path.node) && !path.node.source) {
+    for (const specifier of path.node.specifiers) {
+      if (specifier.exported.name !== specifier.local.name &&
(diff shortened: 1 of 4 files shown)


packages/analyzer/src/model/reference.ts
```diff
@@ -122,6 +122,17 @@ function resolveScopedAt<K extends keyof FeatureKindMap>(
     const exportedIdentifier = getExportedIdentifier(path.node, identifier);
     return resolveThroughImport(path, exportedIdentifier, document, kind);
   }
+
+  // Handle cases like `export {Foo as Bar}`.
+  if (babel.isExportNamedDeclaration(path.node) && !path.node.source) {
+    for (const specifier of path.node.specifiers) {
+      if (specifier.exported.name !== specifier.local.name &&
```
rictic: Why this !== check?

rictic: Talked it out, the other case is handled separately

aomarks: So the way this works is:

1) If `node` is an import from another document, go resolve in that document instead. This logic already its own handling to make sure that in the `foo as bar` case, we look for `foo` in that other document, instead of `bar`.

2) Otherwise, look for a feature with the identifier we're resolving in the local scope of `node`. This is why the `export {foo}` case already worked without any special handling for that syntax (we know `foo` and we know the scope).

I took a little stab at unifying the `bar` -> `foo` logic between the import and local cases, but it didn't seem to be actually more readable without maybe a larger refactoring.

PR #569, opened 2018-07-09
Upgrade typescript to 2.9.2

justinfagnani:


aomarks:
Just had to run `update-goldens` on CLI. In general you can run the top-level `update-goldens` script and that will update all of them.

PR #572, opened 2018-07-10
[gen-typescript-declarations] Add test cases for 3.0 Polymer and elements.

aomarks:
Also support npm install for the fixtures, speed up cloning/installing, and *actually* run the tests on CI. Tests don't run on Windows yet, Analyzer runs out of memory analyzing Polymer for some reason.

We're not producing valid output for modules yet, but this will make the follow up PR diffs easier to read.


rictic:
I believe you can run node with a higher max memory limit, may help the Windows issue. Analyzer is definitely a memory hog at the moment.

aomarks:
> Awesome!



> I believe you can run node with a higher max memory limit, may help the Windows issue. Analyzer is definitely a memory hog at the moment.

Cool, will try it later.

aomarks:
> Awesome!

These aren't useful module typings yet, just in case you missed that :)

PR #574, opened 2018-07-10
[gen-typescript-declarations] Some incremental modules improvements.

aomarks:
- Support serializing export declarations.
- Don't treat ES module imports as HTML imports.
- Rename `update-goldens` script so it runs at top-level.
- Declare `HTMLElementTagNameMap` in global namespace for modules.

### DIFF
--- a/packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/elements/array-selector.d.ts
+++ b/packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/elements/array-selector.d.ts
@@ -186,6 +186,9 @@ declare class ArraySelector extends
   Polymer.Element) {
 }
 
-interface HTMLElementTagNameMap {
-  "array-selector": ArraySelector;
+declare global {
+
+  interface HTMLElementTagNameMap {
+    "array-selector": ArraySelector;
+  }
 }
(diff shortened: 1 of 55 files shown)


packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/elements/array-selector.d.ts
```diff
@@ -191,6 +186,9 @@ declare class ArraySelector extends
   Polymer.Element) {
 }
 
-interface HTMLElementTagNameMap {
-  "array-selector": ArraySelector;
+declare global {
```
rictic: Does this work if the containing file does not contain any imports or exports?

rictic: I think you can safely add an `export {}` line to a file to force it to be a module without otherwise changing semantics

aomarks: Yeah, it does not work unless tsc thinks its a module. Added an `export {}` if there aren't already some exports.

aomarks:
Added a new file that was missing.

PR #581, opened 2018-07-13
[gen-typescript-declarations] Export identifiers.

aomarks:


packages/gen-typescript-declarations/src/test/goldens/paper-behaviors-3/paper-button-behavior.d.ts
```diff
@@ -8,6 +8,8 @@
  *   paper-button-behavior.js
  */
 
+export {PaperButtonBehaviorImpl};
```
rictic: Does this resolve to anything?

aomarks: Yep, only identifiers that can be resolved are included. I'm not sure why we currently export the "Impl" behavior objects, but that's what modulizer did.

aomarks: Oh I suppose you mean does it resolve to a valid TypeScript symbol? No, it doesn't. I'll need to address that later. A lot of funky things going on here too, e.g. the way namespaces are handled.

rictic: It turns out that we use both behaviors and their impl objects in different places in the elements, so that's correct and load bearing.

packages/gen-typescript-declarations/src/es-modules.ts
```diff
@@ -21,3 +21,37 @@ export function isEsModuleDocument(doc: analyzer.Document):
...
+  for (const kind of resolveKinds) {
+    const reference = new analyzer.ScannedReference(
+        kind,
+        identifier,
+        exportFeature.sourceRange,
+        exportFeature.astNode,
+        exportFeature.astNodePath);
+    const resolved = reference.resolve(doc);
+    if (resolved.feature !== undefined) {
+      return resolved as analyzer.Reference<ResolvedTypes>;
```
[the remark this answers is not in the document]
aomarks: Doesn't seem to work, I get `Property 'kind' does not exist on type 'Element'`. What I really wanted is to not have to specify both the kind strings and the type union (`resolveKinds` and `ResolveTypes`) somehow.

Issue #279, opened 2018-05-01
excludes in JS minify being skipped


Westbrook:
I've get a situation very close to https://github.com/Polymer/polymer-cli/issues/701 and while I've applied the approach added via https://github.com/Polymer/polymer-cli/pull/952 I'm noticing that the files in question are still being minified in a way that triggers errors in the Firebase code.

Tracking through the `build`/`cli` code I noticed that when the files are entering the `getOptimizeStreams` section of the build pipeline, the `file` was being tracked with a rewritten `file.pathName` such that `file. relative` would return an unmatchable file name when passed to `notExcluded` for comparison. The can be experienced via https://github.com/Westbrook/polymer-build-filenames which outlines the steps to see this in action.

Upon further review with some help from @justinfagnani I realized that all the the JS files, no matter their origin, were being seen as `isInline === true` at https://github.com/Polymer/tools/blob/master/packages/build/src/html-splitter.ts#L189 which would clearly cause this issue. It seems be happening because `bundle` is occurring before `getOptimizeStreams` here https://github.com/Polymer/tools/blob/master/packages/cli/src/build/build.ts#L84 It seems that this issue can be avoided by moving the `if (bundled) {` block to just before the `if (options.basePath) {` block here https://github.com/Polymer/tools/blob/master/packages/cli/src/build/build.ts#L128 and wanted to see if that made sense to someone more knowledgable on this project that I.

I'd be happy to PR this if you'd like, though I know there is lots of dev going against this right now. Either way let me know if I can help get the right fix for this into the codebase!

davie-robertson:
I thought I was going mad - I've been facing the same issue for weeks.

https://stackoverflow.com/questions/50073467/

Justin - I think this is the same issue (or at least related) I asked you to take a look at (via twitter) regarding a lib  (UnityLoader.js) that keeps getting screwed with in the build process (even taking the code from my custom element and incorporating in directly in my App now breaks during an ES5 build)

Westbrook:
FYI, I'm still experiencing this issue with a fresh install of v1.7.3. I'd be happy to submit a PR if the code ordering was seen as an acceptable fix for this issue. Please advice.

christophe-g:
Gentle ping. 

Seems to be an important issue (... at least preventing me to upgrade).

christophe-g:
@usergenic - any chance for the cli to honor `minify.exclude` as it used to ?

christophe-g:
@usergenic - not giving up ; )

[...]

PR #584, opened 2018-07-13
Allow minify skipping by file glob.

Westbrook:
Allow JS files to be see as `isInline === false` so that they will skip the minifier.

When working with a Polymer 2 project that includes the Firebase and PolymerFire code, the minification processing delivers those projects with errors. This can be avoided by not minifying those files. With the updates to support Polymer 3, all files were being viewed as `isInline` and as such were not being tested for whether they should or shouldn't be minified. This corrects the code order so that files that are not `isInline` can be seen as such and can be tested for exclusion from the minification pipeline.

Fixes https://github.com/Polymer/tools/issues/279

### DIFF
--- a/packages/cli/src/build/build.ts
+++ b/packages/cli/src/build/build.ts
@@ -83,22 +83,6 @@ export async function build(
                 }));
   }
 
-  if (bundled) {
-    // Polymer 1.x and Polymer 2.x deal with relative urls in dom-module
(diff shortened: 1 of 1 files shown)


usergenic:
@Westbrook merging this to a branch and updating goldens and adding regression test.

Westbrook:
🙏 hope I was taking it in a good direction. Thanks for picking this up!

PR #585, opened 2018-07-13
New wct-mocha package

usergenic:
This isn't quite ready to merge.  I just wanted to try this out on CI.


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --ci",
     "build": "lerna run build --stream",
-    "nuke": "rm -rf package-lock.json node_modules && npm install && lerna exec 'rm -f package-lock.json npm-shrinkwrap.json' && lerna clean --yes && lerna exec --stream -- npm install && lerna bootstrap",
```
TimvdLippe: This is necessary to generate the correct lockfiles to work for the `audit` scripts. I don't think we should revert this.

usergenic: @TimvdLippe the only problem right now is that when creating a new package in the monorepo, npm install outside of lerna bootstrap's installation will fail until that package is published to npm.

I'll look at addressing this before proposing the PR for merge, perhaps by preparing a wct-mocha pre-release.

usergenic: @TimvdLippe I wanted to come back to this question because AFAICT the only items not present in the package locks are the packages already in the monorepo, each of which express their own dependencies in their package lock files, so the audit script would catch security issues in those...

usergenic: Let me know if I'm missing something there :)

TimvdLippe: When I run this modified version (e.g. without the `lerna exec --stream -- npm install`) and then I run `npx lerna exec --stream -- npm audit` I receive the following output:

```
lerna notice cli v3.1.4
lerna info versioning independent
cleankill: npm ERR! code EAUDITNOLOCK
cleankill: npm ERR! audit Neither npm-shrinkwrap.json nor package-lock.json found: Cannot audit a project without a lockfile
cleankill: npm ERR! audit Try creating one first with: npm i --package-lock-only
cleankill: npm ERR! A complete log of this run can be found in:
cleankill: npm ERR!     /home/tim/.npm/_logs/2018-08-24T13_58_20_829Z-debug.log
lerna ERR! npm audit exited 1 in 'cleankill'
lerna ERR! npm audit stderr:
npm ERR! code EAUDITNOLOCK
npm ERR! audit Neither npm-shrinkwrap.json nor package-lock.json found: Cannot audit a project without a lockfile
npm ERR! audit Try creating one first with: npm i --package-lock-only

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/tim/.npm/_logs/2018-08-24T13_58_20_829Z-debug.log

lerna ERR! npm audit exited 1 in 'cleankill'
lerna WARN complete Waiting for 3 child processes to exit. CTRL-C to exit immediately.
```

It seems that `lerna bootstrap` does not correctly generate the lock files as required for the `audit` script. Apparently the `audit` script only looks at the lock file and not just the content of `package.json`.

usergenic: Ah interesting.  it looks then like the cause of this deficiency is actually that cleankill doesn't have any dependencies, so lerna doesn't create a package-lock.json for it.  I think I can remedy with a followup to the lerna bootstrap phase.

Thanks for details!

usergenic: @TimvdLippe fyi here's my fix for nuke in a separate PR https://github.com/Polymer/tools/pull/678

Issue #405, opened 2018-05-22
web-component-tester, test fail: Cannot define class using reflection


lemoustachiste:
Hi,

we have a simple project, in its very early days, where we have a basic test for the only webcomponent we have.

Today, I tried running the tests for the first time and encountered a couple of issues I could fix (Safari problem, Java version).

I did install Java v10 and since then have been getting this error:

`Cannot define class using reflection`.

This is the output of `java --version`:

```
java 10.0.1 2018-04-17
Java(TM) SE Runtime Environment 18.3 (build 10.0.1+10)
Java HotSpot(TM) 64-Bit Server VM 18.3 (build 10.0.1+10, mixed mode)
```

Here is the full stack trace:

```{
  "value": {
[...]

PR #586, opened 2018-07-15
Mention the Java 8 dependency in README.md template for new Polymer 3 applications

rjsteinert:
Related to #405. My guess is most people will get stuck like me with tests not running because the directions result in Java 10 being installed when you actually need to install Java 8.


rjsteinert:
@aomarks Good suggestion. I added it.

PR #589, opened 2018-07-17
[gen-typescript-declarations] Emit ESM imports.

aomarks:


packages/gen-typescript-declarations/src/ts-ast/declarations.ts
```diff
@@ -401,10 +401,10 @@ export class ConstValue {
 }
 
 /**
- * An identifier that is imported or exported, possibly with a different name.
+ * An identifier that is imported, possibly with a different name.
  */
-export interface ImportExportIdentifier {
-  identifier: string;
+export interface ImportSpecifier {
+  identifier: string|typeof AllIdentifiers;
```
justinfagnani: cool!

You can also do `type AllIdentifiers = typeof AllIdentifiers`

packages/gen-typescript-declarations/src/ts-ast/declarations.ts
```diff
@@ -413,16 +413,77 @@ export interface ImportExportIdentifier {
...
+      // import, but no named imports.
+      const parts = [];
+      for (const identifier of this.identifiers) {
+        if (identifier.identifier === 'default') {
+          parts.push(identifier.alias);
+        } else if (identifier.identifier === AllIdentifiers) {
+          parts.push(`* as ${identifier.alias}`);
+        }
+      }
+      return `import ${parts.join(', ')} ` +
```
justinfagnani: don't named imports still need to be in `{}`s?

packages/gen-typescript-declarations/src/ts-ast/declarations.ts
aomarks: Good idea, done.

PR #592, opened 2018-07-18
[gen-typescript-declarations] More progress towards Polymer 3.0 typings.

aomarks:
- Adds an `autoImport` config option to automatically add ES module imports when particular identifiers are referenced. We'll use this to inject imports for custom interfaces like `PolymerInit` etc.
- I'm keeping a branch of changes to Polymer 3.0 itself at https://github.com/Polymer/polymer/compare/polymer-3-typescript, you can see the new `interfaces.d.ts` and `gen-tsd.json` files here.
- Automatically detect if a project uses NPM or Bower and configure module resolution settings accordingly.
- Automatically import/export synthetic mixin constructor interfaces.
- Support serializing top-level `const` values with `declare`.
- Use SHAs instead of branches for the 3.0 elements since they might change and break Travis.

### DIFF
--- a/packages/gen-typescript-declarations/src/gen-ts.ts
+++ b/packages/gen-typescript-declarations/src/gen-ts.ts
@@ -64,6 +64,15 @@ export interface Config {
    * e.g. entire generated classes.
    */
   renameTypes?: {[name: string]: string};
+
+  /**
+   * A map from an ES module path (relative to the analysis root directory) to
+   * an array of identifiers exported by that module. If any of those
+   * identifiers are encountered in a generated typings file, an import for that
+   * identifier from the specified module will be inserted into the typings
+   * file.
+   */
+  autoImport?: {[modulePath: string]: string[]};
 }
 
 const defaultExclude = [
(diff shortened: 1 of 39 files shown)


packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -463,17 +536,43 @@ class TypeGenerator {
       ],
       returns: new ts.IntersectionType([
         new ts.NameType('T'),
-        new ts.NameType(mixinName + 'Constructor'),
-        ...Array.from(this.transitiveMixins(feature))
-            .map((mixin) => new ts.NameType(mixin + 'Constructor'))
+        new ts.NameType(constructorName),
+        ...[...transitiveMixins].map(
+            (mixin) =>
+                new ts.NameType([...mixin.identifiers][0] + 'Constructor'))
```
justinfagnani: nit: I like putting expressions like `[...mixin.identifiers][0] + 'Constructor')` into a named variable or function to make it more self documenting. Since you use this below too, it'll document both cases. And would it be used for `constructorName` as well? (why is one using `feature.name` and the other the first identifier?)

aomarks: I just forgot that `.name` was a thing, switched to that.

PR #594, opened 2018-07-19
Hoist `@types/node`

usergenic:
Since this was a dependency (i.e. not devDependency) of some packages, I updated the version only in those packages.  Wherever it was a devDependency, I removed it after hoisting it to top-level.

Note I also had to update two source files because of the typings issue with `AddressInfo|string` of the `Server#address` property.


justinfagnani:
If there any rationale for it being a dep vs devdep?

usergenic:
I was going to just say I didn't assess those occurrences where it was defined as a `dependency` and leave it at that.  On further thought though, it seems that most of the time we make types a non-dev-dependency if our package's public interface transacts in values covered by those types.

I mean, I would evaluate the dev vs non-dev decision in those terms, but in the case of node typings specifically, maybe making them published prod dependencies isn't the way to go, since the correct typings are going to be more a reflection of the version of node that is being built for?

I definitely want to get feedback from team before I pull typings out of `dependencies`, especially for `polymer-analyzer` which is one of our more frequently used-as-a-library packages.

WDYT?

packages/cleankill/package.json
```diff
@@ -14,6 +14,6 @@
   "typings": "./index.d.ts",
   "engine": ">=4",
   "dependencies": {
-    "@types/node": "^8.10.19"
+    "@types/node": "^10.5.2"
```
justinfagnani: This can be a devDep

packages/common/package.json
```diff
@@ -19,7 +19,7 @@
     "format": "find src/ test/ -iname '*.ts' -o -iname '*.js' | xargs clang-format --style=file -i"
   },
   "dependencies": {
-    "@types/node": "^9.6.4",
+    "@types/node": "^10.5.2",
```
justinfagnani: I think common doesn't need this?

packages/editor-service/package.json
```diff
@@ -32,7 +32,7 @@
   },
   "dependencies": {
     "@types/minimatch": "^3.0.1",
-    "@types/node": "^9.6.4",
+    "@types/node": "^10.5.2",
```
justinfagnani: I bet this can be a devDep

packages/linter/package.json
```diff
@@ -30,7 +30,7 @@
     "@types/babel-traverse": "^6.25.2",
     "@types/babel-types": "^6.25.1",
     "@types/clone": "^0.1.30",
-    "@types/node": "^8.10.11",
+    "@types/node": "^10.5.2",
```
justinfagnani: devDep?

packages/project-config/package.json
```diff
@@ -16,7 +16,7 @@
     "format": "find src test \\( -iname \"*.ts\" -o -iname \"*.js\" \\) | xargs clang-format --style=file -i"
   },
   "dependencies": {
-    "@types/node": "^9.6.4",
+    "@types/node": "^10.5.2",
```
justinfagnani: devDep?

PR #596, opened 2018-07-19
Updated package locks and goldens.

usergenic:
I was working on another branch that needed a good old nuking and updated babel stuff which generates different output needing updated goldens and I thought I'd separate that out into this separate PR where I just `npm run nuke` and `npm run update-goldens` for review.  So here is that PR.


usergenic:
I'm having a side-thought which is, do we have enough coverage on non-minified build output so we can ensure updated babel compilers are doing the right thing before the output gets transformed into `¯\_(ツ)_/¯`...

usergenic:
Obviated by https://github.com/Polymer/tools/pull/594

Issue #597, opened 2018-07-20
Fix flaky integration tests; retry?


usergenic:
![screen shot 2018-07-20 at 12 26 47 pm](https://user-images.githubusercontent.com/578/43021401-60a5689e-8c18-11e8-8a3c-ab95a99a513e.png)

These most frequently bomb our `web-component-tester` integration tests, (most frequently when testing on sauce, but not necessarily) which ends the integration test run after wct, so that we don't get cli tests run at all and we either ignore or endlessly rerun the whole integration test battery on Travis in the hopes we can get to green.

This is problematic when Travis *never* gets to green.  I've been trying to get to green on https://github.com/Polymer/tools/pull/596 to prove out the updates to babel minify are compatible with all the browsers tested on sauce, but the test runs are plagued by timeouts and rerunning them still doesn't get to the finish line without one or more wct test timeouts.

Really want to get integration tests to retry on failure up to 3 times or something.  This will require redefining how wct's integration tests are structured, since the retry really needs to happen on the setup that occurs in a mocha `before` right now which does not support retry.

usergenic:
It turns out I *did* fix this a while back by creating a retry-failures-reporter for mocha.  Here's the last PR that fixed issues with it for reference. https://github.com/Polymer/tools/pull/739

PR #598, opened 2018-07-20
Make WCT integration tests retry on errors

usergenic:
These tests are flaky, especially when Sauce is involved, but even just locally.  Until we can make them more reliable, we need to at least get them to retry after a failure so that we can have a shot at a green test run instead of having WCT integrations fail so much we ignore them.


usergenic:
Fixes https://github.com/Polymer/tools/issues/597

usergenic:
So... it turns out retrying on these kind of errors actually only make things worse in the typical case, since the errors themselves are generally not transient hiccups, but service availability and capacity issues.

PR #599, opened 2018-07-21
[gen-typescript-declarations + analyzer] Better class heritage handling.

aomarks:
The Polymer 3 (and 2) types are complete AFAIK and compiling cleanly after these changes. These last changes are mostly around handling class heritage more correctly. 

**Analyzer changes**

1) Parse `@extends {Foo}` annotations (previously only `@extends Foo` worked).

2) Treat `@constructor` annotations as classes. This is helpful for understanding class heritage and fixes super classes for some of the bundled Polymer elements. It also introduces some noise, so I added some `@private` annotations in Polymer (I also found some regular class declarations that should have been `@private` while I was at it).

3) Add an `isLegacyPolymerFactoryCall` property to `PolymerElement` features so that we can do some special casing there. Previously there was no good way to distinguish between calls to `Polymer()` and any other kind of custom element (even non-Polymer CEs are considered `PolymerElements` by Analyzer).

4) Collapse "ephemeral" super classes into their children. This is to handle the pattern where a class's superclass is declared in a separate variable just to satisfy the Closure compiler. We no longer emit these classes, and instead merge their super class and mixin information into the children.

**Type generator changes**

1) Handle super classes more correctly. Previously we just did it kind of wrong and assumed too many things extended `PolymerElement`. Now we simply trust Analyzer's model, and have a special case for legacy Polymer function calls, which now extend `PolymerLegacyMixin & HTMLElement` instead of `PolymerElement`.

2) Mixin instance interfaces now extend the instance interfaces for the mixins that they automatically apply. This worked before when you received an interface via a constructor call, because the constructor hierarchy was complete, and TypeScript is smart enough to figure out the instance hierarchy from that. But this way you can also cast directly to the interface and get the full picture.

**Polymer source changes**

- [Polymer 3](https://github.com/Polymer/polymer/compare/polymer-3-typescript)
- [Polymer 2](https://github.com/Polymer/polymer/compare/2.x...2.x-typescript)


aomarks:
The CLI integration test is failing because linter is now picking up the [double definition of Polymer.Element in 2.x](https://github.com/Polymer/polymer/blob/2.x/polymer-element.html). My [pending changes](https://github.com/Polymer/polymer/compare/2.x...2.x-typescript) to the 2.x branch fix this, so if the changes look good we would need to tag a 2.x release to make the CLI tests pass again.

packages/analyzer/src/javascript/class-scanner.ts
```diff
@@ -182,6 +186,62 @@ export class ClassScanner implements JavaScriptScanner {
...
+      }
+      if (superClass.privacy !== 'private') {
+        continue;
+      }
+      let isEphemeral = false;
+      for (const childClass of childClasses) {
+        if (superClass.superClass !== undefined) {
+          // Feature properties are readonly, hence this hacky cast. We could
+          // also make a new feature, but then we'd need a good way to clone a
+          // feature.
```
rictic: Maybe mention that this is ok because this is still part of the "construction" phase of the scanned classes, they're entirely file local, and they haven't left the ClassScanner yet.

packages/analyzer/src/polymer/polymer-element.ts
```diff
@@ -150,14 +150,14 @@ export interface Options {
...
   behaviors: ScannedReference<'behavior'>[];
-
   events: Map<string, ScannedEvent>;
-
   abstract: boolean;
   privacy: Privacy;
   astNode: AstNodeWithLanguage;
   statementAst: babel.Statement|undefined;
   sourceRange: SourceRange|undefined;
+  /** Whether this element was defined using the legacy Polymer() function. */
```
rictic: Nit, slightly clearer as: `true iff this element was defined using the legacy Polymer() function.`

packages/gen-typescript-declarations/scripts/update-goldens.js
```diff
@@ -30,10 +30,18 @@ const {generateDeclarations} = require('../lib/gen-ts');
 const fixturesDir = path.join(__dirname, '..', 'src', 'test', 'fixtures');
 const goldensDir = path.join(__dirname, '..', 'src', 'test', 'goldens');
 
-fsExtra.emptyDirSync(goldensDir);
+const filter = process.env.UPDATE_GOLDENS_FILTER || '';
```
rictic: Why add the filter?

packages/analyzer/src/javascript/class-scanner.ts
```diff
@@ -182,6 +186,62 @@ export class ClassScanner implements JavaScriptScanner {
...
+      }
+    }
+
+    const ephemeralSuperClasses = new Set<ClassType>();
+    for (const [superClassId, childClasses] of classBySuperClassId) {
+      const superClass = classById.get(superClassId);
+      if (superClass === undefined) {
+        continue;
+      }
+      if (superClass.privacy !== 'private') {
```
rictic: We should also check that it's an assignment to a local variable.

packages/analyzer/src/javascript/class-scanner.ts
aomarks: Done.

packages/analyzer/src/polymer/polymer-element.ts
aomarks: Done.

packages/gen-typescript-declarations/scripts/update-goldens.js
aomarks: It's to make it easy to work on one of the test cases at a time with a faster development loop, since it takes a few seconds per case.

packages/analyzer/src/javascript/class-scanner.ts
aomarks: Done, and added a comment for the requirements of an ephemeral superclass.

aomarks:
I've temporarily disabled the lint check for the Polymer 2.x application template because it will return a warning until https://github.com/Polymer/polymer/pull/5300 is released to bower. That change depends on this PR being released to the generator, though, so I'll re-enable this test afterwards.

PR #604, opened 2018-07-23
Run a separate VM for sauce-oriented integration tests.

usergenic:
Rerunning the whole test suite every time WCT on sauce has a sad is a bad experience because it takes 10 minutes to even get there.  By running `web-component-tester` and `@polymer/esm-amd-loader` tests separately we have a shorter path to rerunning WCT when it flakes.


aomarks:
Rebased this to master, since there was a unit test failure that should now be fixed. Looks like the integration tests were passing after this PR so that's promising.

aomarks:
Looks like this is passing, @usergenic was this ready for review?

PR #610, opened 2018-07-30
Updated the list of browsers to run sauce tests for.

usergenic:
Adding coverage for latest Safari and Edge to sauce browser list.


usergenic:
Closing PR in favor or #615

PR #612, opened 2018-07-31
cli: escape chalk parentheses

### LINKED ISSUE #595: [polymer-cli] Error: Found extraneous } in Chalk template literal
Using polymer-cli in version `1.7.7`, the following command:

```
polymer help lint
```

produces an error:

```
error: [cli.main]   cli runtime exception: Error: Found extraneous } in Chalk template literal
error: [cli.main]   Error: Found extraneous } in Chalk template literal
    at tmp.replace (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/node_modules/chalk/templates.js:109:11)
    at String.replace (<anonymous>)
    at module.exports (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/node_modules/chalk/templates.js:99:6)
    at chalkTag (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/node_modules/chalk/index.js:221:9)
    at Chalk.chalk.template (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/node_modules/chalk/index.js:36:20)
    at format (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/lib/chalk-format.js:4:10)
    at chalkFormat (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/lib/chalk-format.js:10:12)
    at ContentSection.arrayify.map.line (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/lib/content-section.js:14:63)
    at Array.map (<anonymous>)
    at new ContentSection (/home/lpellegr/.nvm/versions/node/v10.6.0/lib/node_modules/polymer-cli/node_modules/command-line-usage/lib/content-section.js:14:51)
```

43081j:
So this seems to fix #595.

---

The underlying issue here is that command-line-usage uses chalk 2.x (we use chalk 1.x) which makes heavy use of template literals and introduces a curly-bracket syntax like so:

```
{bold some bold text} some normal text
```

When we pass our help text to command-line-usage, it isn't escaped (because they forcefully pass it to chalk as "raw" text which chalk doesn't escape). So our parentheses are interpreted as chalk's syntax.

Internally, chalk escapes parentheses with the exact regex i've added here.

I'm not sure if i like this fix... because this could easily occur in other places where we pass arbitrary text into chalk. Any ideas how we can generalise this a little more? Or if we need to?


packages/cli/src/commands/lint.ts
```diff
@@ -128,7 +128,11 @@ export class LintCommand implements Command {
         content: collectionsDocs.join('\n\n'),
         raw: true
       },
-      {header: 'Lint Rules', content: rulesDocs.join('\n\n'), raw: true}
+      {
+        header: 'Lint Rules',
+        content: rulesDocs.join('\n\n').replace(/[{}\\]/g, '\\$&'),
```
aomarks: Can you add a comment to explain this?

43081j:
@aomarks  thanks for taking a look at this, much appreciated.

have updated the changelog and added a small comment now

Issue #620, opened 2018-08-05
[modulizer] Trim whitespace for user inputs


greenyouse:
When running the modulizer it prompts for a npm package name and version number. Currently whitespaces aren't being trimmed from the input so the extra spaces end up in the generated `package.json` fields. 

Could these strings be trimmed to handle the issue?

PR #621, opened 2018-08-05
[modulizer]  Trim whitespaces for user input

greenyouse:
A little tweak to trim whitespaces. I didn't see any tests for this part so none were added.

This is a fix for #620


TimvdLippe:
I think you can base your tests off https://github.com/Polymer/tools/blob/4911035bee8defc5a2fd7dcbc04025e9c852b646/packages/cli/src/test/integration/integration_test.ts#L145-L153

PR #624, opened 2018-08-06
cli: remove obsolete types

43081j:
I bumped into these during my other cli PRs.

@rictic converted these two repos to typescript some time ago afaik, so we don't need these custom types anymore.

maybe we could also scrap the command-line-* types and pull in the `@types` ones?


usergenic:
Thanks for your PR @43081j

PR #631, opened 2018-08-10
[gen-typescript-declarations] Fixes for 3.0 elements.

aomarks:
- Remove `_template` property from legacy Polymer function components.
- Functions defined within exported objects are no longer top-level functions.
- Remove broken imports/exports for `FooBehaviorImpl` style behaviors.
- Update fixture SHAs.
- Update URL for this repo put into generated file headers.

cc https://github.com/Polymer/gen-typescript-declarations/issues/79 @TimvdLippe @43081j


packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -1037,6 +1041,29 @@ function isPolymerElement(feature: analyzer.Feature):
...
+ *   export const FooBehavior = [FooBehaviorImpl, OtherBehavior];
+ *
+ * In this case, Analyzer merges FooBehaviorImpl into FooBehavior and does not
+ * emit a behavior feature for FooBehaviorImpl. However, there is still an
+ * export feature for FooBehaviorImpl, so we exclude it here.
+ */
+function isBehaviorImpl(reference: analyzer.Reference<analyzer.Feature>) {
+  return reference.feature !== undefined &&
+      reference.feature.kinds.has('behavior') &&
+      (reference.feature as analyzer.PolymerBehavior).name !==
```
43081j: This made me wonder, could `kinds` be a `Set<keyof FeatureKindMap>` instead of `Set<string>`? means if we do this pattern often we could just have a `feature is T` function and have the correct typing. maybe? unrelated to this PR though.

packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/legacy/templatizer-behavior.d.ts
```diff
@@ -114,37 +114,3 @@ interface Templatizer {
 }
 
 declare const Templatizer: object;
```
TimvdLippe: This removes a lot of the type information that we previously had. Can we be smarter here and instead export the proper object type with the function definitions as they were previously below it?

packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/legacy/polymer.dom.d.ts
```diff
@@ -35,6 +35,20 @@ declare class DomApi {
    * shadow root.
    */
   readonly activeElement: Node|null|undefined;
+  parentNode: Node|null;
```
TimvdLippe: Nice!

packages/gen-typescript-declarations/src/test/goldens/polymer-3/lib/legacy/templatizer-behavior.d.ts
aomarks: I think this is actually correct, and it's the same as what we have always done for behavior typings. For behaviors, we export two things with the same name: 1) an interface which represents how the behavior will augment a class, and 2) a plain object which is just the thing you pass to the behaviors array or the mixinBehaviors function. The behavior object itself shouldn't have methods on it, it's more like the constructor than the prototype.

Before this PR we were also incorrectly declaring all of the functions of the behavior as duplicated top-level functions, that's the only change.

packages/gen-typescript-declarations/src/gen-ts.ts
aomarks: Maybe, would be nice to make these type guards easier.

PR #632, opened 2018-08-10
[gen-typescript-declarations] Add --verify command-line flag.

aomarks:
Runs the generated typings through the TypeScript compiler (3.0) and fails with its error log if it fails.

The goal is to make integrating fully tested type generation into new repos (the Polymer elements) as trivial as possible.

cc https://github.com/Polymer/gen-typescript-declarations/issues/79 @43081j

### DIFF
--- a/packages/gen-typescript-declarations/src/cli.ts
+++ b/packages/gen-typescript-declarations/src/cli.ts
@@ -14,6 +14,7 @@ import * as glob from 'glob';
 import * as path from 'path';
 
 import {Config, generateDeclarations} from './gen-ts';
+import {verifyTypings} from './verify';
 
 import commandLineArgs = require('command-line-args');
 import commandLineUsage = require('command-line-usage');
@@ -53,6 +54,12 @@ const argDefs = [
         'writing new typings, excluding node_modules/, bower_components/, ' +
         'or any file added using the <addReferences> config option.',
   },
+  {
+    name: 'verify',
+    type: Boolean,
+    description: 'Compile the generated types with TypeScript and fail if ' +
+        'they are invalid.',
+  },
 ];
 
 interface Args {
@@ -62,6 +69,7 @@ interface Args {
   config?: string;
   outDir?: string;
   deleteExisting?: boolean;
+  verify?: boolean;
 }
 
 async function run(argv: string[]) {
@@ -90,6 +98,7 @@ async function run(argv: string[]) {
   if (!args.outDir) {
     throw new Error('--outDir is required');
   }
+  const outDir = path.resolve(args.outDir);
 
   if (!args.config) {
     const p = path.join(args.root, 'gen-tsd.json');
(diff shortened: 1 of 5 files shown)


packages/gen-typescript-declarations/src/verify.ts
```diff
@@ -0,0 +1,62 @@
...
+    };
+  }
+  return {success: true, errorLog: ''};
+}
+
+/**
+ * The result of calling verifyTypings.
+ */
+export interface VerifyTypingsResult {
+  success: boolean;
```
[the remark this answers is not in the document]
aomarks: Seems like an unnecessary cross-package dependency for such a trivial type.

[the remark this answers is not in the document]
43081j: Yup I had assumed it already used the analyser elsewhere. My mistake 🙈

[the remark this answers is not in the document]
aomarks: Oh the package does, so it wouldn't introduce a new npm dependency, it's more just a style preference to avoid unnecessary module level dependencies as a rule.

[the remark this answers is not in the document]
43081j: ah! yup i understand now :)

packages/gen-typescript-declarations/src/cli.ts
```diff
@@ -53,6 +54,12 @@ const argDefs = [
         'writing new typings, excluding node_modules/, bower_components/, ' +
         'or any file added using the <addReferences> config option.',
   },
+  {
+    name: 'verify',
+    type: Boolean,
+    description: 'Compile the generated types with TypeScript and fail if ' +
```
TimvdLippe: Lets add the version in the description to be super clear and prevent confusion later.

packages/gen-typescript-declarations/src/verify.ts
```diff
@@ -0,0 +1,62 @@
...
+  errorLog: string;
+}
+
+const compilerOptions = {
```
TimvdLippe: Can we assume the options here? Maybe also support reading in the tsconfig.json if it exists, so that we use the user configuration if possible.

TimvdLippe: Nit: move the options before the usage in the function.

aomarks: > Can we assume the options here? Maybe also support reading in the tsconfig.json if it exists, so that we use the user configuration if possible.

I think it's a pretty good assumption. In a real package, it doesn't matter what tsconfig.json the author used, only the end-users one is used. So ensuring that anything generated here would work with what we think is a good tsconfig seems pretty reasonable. Could always add some override ability later.

> Nit: move the options before the usage in the function.

Done.

packages/gen-typescript-declarations/src/cli.ts
aomarks: Done.

PR #636, opened 2018-08-12
[gen-typescript-declarations] Make autoImport also work with parameter types.

aomarks:


packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -231,7 +231,7 @@ function addAutoImports(tsDoc: ts.Document, autoImport: Map<string, string>) {
   const alreadyImported = getImportedIdentifiers(tsDoc);
 
   for (const node of tsDoc.traverse()) {
-    if (node.kind !== 'name') {
+    if (node.kind !== 'name' && node.kind !== 'param') {
```
rictic: How does this work? Why does skipping params mean that they get autoimported?

aomarks: It's negated, so it's included, not skipped, I'll rewrite it to be less confusing.

aomarks: Done.

PR #639, opened 2018-08-14
[gen-typescript-declarations] Show analyzer warnings when generating types

aomarks:


aomarks:
Removed the original change that exited if there were analyzer errors. Turns out Polymer core has a bunch of errors, which we should fix, but for now just going to show the errors.

PR #644, opened 2018-08-14
Hoist clang-format package and use a single .clang-format config in root

usergenic:
There's no good reason I can think of to have these all formatting rules be scattered across the monorepo's packages, so this PR standardizes all of them.

I hoisted clang-format package as well as .clang-format file to repo root, then I added format scripts to all packages that didn't have them and normalized those too.  Then I ran format script for all packages, ran build, and committed what you see here.


usergenic:
note: I did this because I was about to add yet-another-package with yet-another-clang-format config and dependency and that felt sad.

Also, this configuration makes it easier to configure editor to use the right clang-format settings because it can pull them from the workspace root instead of figuring it out per-package.

PR #645, opened 2018-08-15
Force "foo" in "import * as foo" to be resolvable.

aomarks:
This is so that we don't need https://github.com/PolymerElements/iron-dropdown/pull/165


aomarks:
> Add a test?

Done, and there was a bug.

(It's kind of confusing that analyzer seems to prefer resolving with exported names over local names, since the exported names aren't really resolvable within the module, they only exist for importers.)

rictic:
I'm not 100% sure what you mean by "prefers resolving by"

aomarks:
> I'm not 100% sure what you mean by "prefers resolving by"

If you have an `export {foo as bar};` export feature, then using the approach I use in `resolveImportExportFeature` (https://github.com/Polymer/tools/blob/master/packages/gen-typescript-declarations/src/es-modules.ts), I believe resolving `bar` will work, but not `foo`. I guess it makes sense in the context of the export, but it has thrown me a few times (I expect to be resolving the local name, since my starting point is the module that is doing the exporting).

PR #652, opened 2018-08-16
[browserCapabilities] Corrected browserName for Opera

### LINKED ISSUE #640: [browserCapabilities] Opera always gets 'es5-bundled'
I noticed that despite that Opera supports es2015 and push it always loads es5-bunled.
Example: https://shop.polymer-project.org

I found that the reason for that is in browserCapabilities.

In `browser-capabilities.js` there is a part with switch-case:

```javascript
    'Chrome': chrome,
    'Chromium': chrome,
    'Chrome Headless': chrome,
    'OPR': {
        es2015: since(36),
        es2016: since(45),
        es2017: since(45),
        es2018: since(51),
        push: since(28),
        serviceworker: since(32),
        modules: since(48),
    },
```

Here `OPR` has to be replaced with `Opera` cause ua-parser-js never returns 'OPR' as broserName for any version of Opera.

P.S. I've tested on Opera 54.

exunreal:
Fixes #640
Changes browserName of Opera in browserCapabilities.

### DIFF
--- a/packages/browser-capabilities/src/browser-capabilities.ts
+++ b/packages/browser-capabilities/src/browser-capabilities.ts
@@ -51,7 +51,7 @@ const browserPredicates: {
   'Chrome': chrome,
   'Chromium': chrome,
   'Chrome Headless': chrome,
-  'OPR': {
+  'Opera': {
     es2015: since(36),
     es2016: since(45),
     es2017: since(45),


aomarks:
Thanks for the PR. Released as 1.1.2.

PR #661, opened 2018-08-21
[gen-typescript-declarations] Expanded TemplateStringsArray type for html function in polymer

SergeShkurko:
Before:
image

After:
image

Unfortunately, the function
```typescript
html`<style>${style}</style>${view()}`
````

Not work if vars style or view loaded from webpack
image
Gives error:
image

### This change solves the problem

### DIFF
--- a/packages/gen-typescript-declarations/src/closure-types.ts
+++ b/packages/gen-typescript-declarations/src/closure-types.ts
@@ -194,7 +194,7 @@ const renameMap = new Map<string, string>([
   // will concede that technicality.)
   ['Object', 'object'],
   // The tagged template literal function argument.
-  ['ITemplateArray', 'TemplateStringsArray'],
+  ['ITemplateArray', 'TemplateStringsArray|string[]'],
 ]);
 
 /*
--- a/packages/gen-typescript-declarations/src/test/goldens/polymer-2/lib/utils/html-tag.d.ts
+++ b/packages/gen-typescript-declarations/src/test/goldens/polymer-2/lib/utils/html-tag.d.ts
@@ -60,7 +60,7 @@ declare namespace Polymer {
    *
    * @returns Constructed HTMLTemplateElement
    */
-  function html(strings: TemplateStringsArray, ...values: any[]): HTMLTemplateElement;
+  function html(strings: TemplateStringsArray|string[], ...values: any[]): HTMLTemplateElement;
 
 
   /**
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by aomarks** [comment]
I think you should probably be using the [`htmlLiteral`](https://github.com/Polymer/polymer/blob/c9b539fc76f88bc6646ceba6fb85c3bf8fdf6877/lib/utils/html-tag.js#L125) function for this use case, like this:

```js
import {PolymerElement, html} from '@polymer/polymer/polymer-element.js';
import {htmlLiteral} from '@polymer/polymer/lib/utils/html-tag.js';

const style = htmlLiteral`
  p { color: red; }
`;

const view = htmlLiteral`
  <p>Hello [[somebody]]!</p>
`;

class MyElement extends PolymerElement {
  static get template() {
    return html`
      <style>${style}</style>
      ${view}
    `;
  }

  static get properties() {
    return {
      somebody: {
        type: String,
        value: 'World',
      },
    };
  }
}

customElements.define('my-element', MyElement);
```

SergeShkurko:
@aomarks this example not solve problem separate styles & html in external files (And use of preprocessors sass, pug etc)

aomarks:
I see. I suppose it is valid to directly invoke template tag literal functions with strings in this way. @justinfagnani any thoughts on this pattern?

However, in any case, I think the correct place to fix this would by editing the actual source file at https://github.com/Polymer/polymer/blob/master/lib/utils/html-tag.js and changing the `@param` annotations to `{!ITemplateArray|!Array<string>}`, because this issue is not specific to the way we transform Closure types to TypeScript types.

aomarks:
Talked this over with @justinfagnani, we actually actively want to discourage this way of invoking the template tag literal functions, because it makes a little too easy to bypass the security provided by the requirement of passing only safe strings tagged by `htmlLiteral`. Ideally you would be able to generate a tagged string literal from your template generator (like the snippet I sent before), but since I assume that's not really possible here, you can simply cast like this:

```ts
html([`<style>${style}</style>`] as TemplateStringsArray)
```

kevinpschaaf:
Cross posting from my comment on #5349:

The `html` tag function uses `innerHTML` to generate a template, and is thus not safe to interpolate untrusted strings. For your use case (interpolating strings from an import), the thing you import needs to have been generated using `html` or `htmlLiteral`; perhaps this could be done by passing your pug file through a webpack transform that wraps with html.  Calling the `html` tag function as a function rather than by tagging strings is considered an abuse of the API, which may break in the future.

We will not support interpolation of arbitrary strings (i.e., not verifiably trusted strings) into Polymer's html tag function, since we want the default that ships with Polymer to be verifiably safe.

Note however, it is not a requirement that you use the html tag function that ships with Polymer in your static template getter. The only requirement is that it return an `HTMLTemplateElement`. You can use your own javascript to make a template however you like, and it should be trivial for you to create a function that innerHTML's a template from whatever sources you want -- we're just not going to write that code for you since it's you (the author) who needs to verify it's safe for your use case. Hope that helps.

PR #662, opened 2018-08-21
npm run nuke + plylog 1.0.0 update

usergenic:
Needed to update stuff, including plylog.  Here is that story in PR form.


package.json
```diff
@@ -15,10 +15,10 @@
     "test": "npm run build && npm run test:unit",
     "test:all": "npm run test:unit && npm run test:integration",
     "test:all:parallel": "npm run test:unit && npm run test:integration:parallel",
-    "test:integration": "lerna run test:integration --stream",
+    "test:integration": "lerna run test:integration --stream --concurrency 1",
     "test:integration:parallel": "lerna run test:integration --stream --parallel",
     "test:integration:windows": "lerna run test:integration --stream --ignore web-component-tester --ignore @polymer/esm-amd-loader --parallel",
-    "test:unit": "lerna run test:unit --stream",
+    "test:unit": "lerna run test:unit --stream --concurrency 1",
```
rictic: how does this affect our testing performance?

packages/esm-amd-loader/test/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "devDependencies": {
     "@polymer/esm-amd-loader": "^1.0.0",
-    "@webcomponents/webcomponentsjs": "^1.2.0",
+    "@webcomponents/webcomponentsjs": "^1.2.4",
```
rictic: Did you file a bug against webcomponentsjs?

rictic: Ah, looks like an IE bug was fixed in 1.2.6

package.json
usergenic: Unit test time impact is negligible and it simplifies output.  If higher concurrency we got a real bump in test time on travis without impacting reliability I'd bump it up.

usergenic: (by negligible I mean not consistently more-than it was.)

PR #672, opened 2018-08-25
Export generated legacy Polymer function element interfaces.

aomarks:
Kind of an important omission I just noticed. Users should be able to import the interface we generate for legacy Polymer elements, even though those element definitions are not themselves exported.

Also prepare 1.5.1 for release.

cc @bicknellr I just verified that all types are still compiling after this change. Once it's released, refreshing all the package-lock.json files is all that's needed to pull in this fix.


rictic:
+1. I added something similar in my goog branch, which I'll to send out for review next week

PR #673, opened 2018-08-26
Add option to generate output for goog.module-compatible modules

rictic:
Hopefully we'll only need this for a short period of time.

* Added `--googModules` flag, for generating declarations compatible
  with `goog.module`-style modules.
* Added `--hideWarnings` flag, for hiding warning-level severity diagnostics.
* Generate declarations with tslint disables for relevant rules, like variable
  names, and no-any.

I separated out the content commits from the commits that update goldens for clarity. Will squash merge after review.


packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -74,6 +74,20 @@ export interface Config {
...
+   * resolved for TypeScript inside google3. This is probably not at all useful
+   * for anyone but the Polymer team.
+   */
+  googModules?: boolean;
+
+  /**
+   * If true, does not log warnings detected when analyzing code,
+   * only diagnostics of Error severity.
+   */
+  hideWarnings?: boolean;
```
TimvdLippe: Instead of a boolean, would it be nice to have the minimum error level here instead? E.g. the `true` now would mean `ERROR`.

packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -176,6 +190,10 @@ async function analyzerToAst(
       header: makeHeader(
           analyzerDocs.map((d) => analyzerUrlToRelativePath(d.url, rootDir))
               .filter((url): url is string => url !== undefined)),
+      tsLintDisables: [{
+        ruleName: 'variable-name',
+        why: `API description`,
```
TimvdLippe: I am not really sure what you mean with this description. Could you clarify?

packages/gen-typescript-declarations/src/gen-ts.ts
rictic: Maybe? It seems like the two reasonable options are either errors only, or errors and warnings, so it would make the API more awkward but no more expressive.

packages/gen-typescript-declarations/src/gen-ts.ts
rictic: Clarified, PTAL.

rictic:
Updated, and switched to the latest goog module id style. PTAL!

packages/gen-typescript-declarations/src/gen-ts.ts
```diff
@@ -272,6 +297,81 @@ function addAutoImports(tsDoc: ts.Document, autoImport: Map<string, string>) {
...
+  let packageInfo: {name?: string};
+  try {
+    packageInfo = JSON.parse(
+        fsExtra.readFileSync(path.join(rootDir, 'package.json'), 'utf-8'));
+  } catch {
+    return undefined;
+  }
+  return packageInfo.name;
+}
+function googModuleForNameBasedImportSpecifier(spec: string) {
```
aomarks: newline

packages/gen-typescript-declarations/src/test/gen-ts_goog_test.ts
```diff
@@ -0,0 +1,50 @@
...
+import * as fs from 'fs';
+import * as glob from 'glob';
+import * as path from 'path';
+
+import {Config, generateDeclarations} from '../gen-ts';
+
+const fixtures = path.join(__dirname, '..', '..', 'src', 'test', 'fixtures');
+const goldens = path.join(__dirname, '..', '..', 'src', 'test', 'goldens_goog');
+
+suite('generateDeclarations (goog)', () => {
```
aomarks: this is nearly all duplicated, why not just update the main test to run the test twice with two different configs?

packages/gen-typescript-declarations/scripts/update-goldens.js
```diff
@@ -29,32 +29,70 @@ const {generateDeclarations} = require('../lib/gen-ts');
...
+    })());
   }
-  console.log('making goldens for ' + fixture);
-  fsExtra.emptyDirSync(path.join(goldensDir, fixture));
 
-  let config = {};
-  const configPath = path.join(fixturesDir, fixture, 'gen-tsd.json');
-  if (fs.existsSync(configPath)) {
-    config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
+  for (const fixture of fs.readdirSync(fixturesDir)) {
```
aomarks: Same here, could just have run the generation/saving code twice instead of copying the whole loop.

packages/gen-typescript-declarations/src/gen-ts.ts
rictic: Done

packages/gen-typescript-declarations/src/test/gen-ts_goog_test.ts
rictic: The choice was deliberate, and guided the maxims: "If you do it twice, copy-paste, if you do it three times, abstract" and "test code should optimize for legibility even more than most".

That said, I'm not dead set on it

packages/gen-typescript-declarations/scripts/update-goldens.js
rictic: See above

PR #674, opened 2018-08-28
analyzer: support mjs extension for js parser

### LINKED ISSUE #671: [polymer-cli] No parser for for file type `mjs`
### Description
Google recommends using `.mjs` files to maintain parity with NodeJS, but `polymer-cli` apparently doesn't recognise them as javascript. When running `polymer-lint`, I got the following error:

```sh
$ polymer lint


  <script type="module" src="src/chatter-app.mjs"></script>
                            ~~~~~~~~~~~~~~~~~~~~~

index.html(66,29) error [could-not-load] - Unable to load import: No parser for for file type mjs


Found 1 error.
```

Also, I think that that error contains a typo.

### Versions & Environment

- Polymer CLI: `1.8.0`
- node: `v10.9.0`
- Operating System: `MacOS Sierra 10.12.6`

#### Steps to Reproduce
Example:

1. Create an application project: `polymer init application`
2. Rename all javascript files to use the `.mjs` convention
3. Change `import`s and `<script>` tags accordingly
3. Lint: `polymer lint`

#### Expected Results
No error is thrown

#### Actual Results
Error is throne: `index.html(66,29) error [could-not-load] - Unable to load import: No parser for for file type mjs`

43081j:
Fixes #671.

I _think_ this is all we need.


aomarks:
Add a note to the changelogs?

aomarks:
Have you manually tested this E2E with polymer cli? Would be nice to actually have an integration test fixture, just in case there are other spots that assume the file extension.

43081j:
@aomarks  sorry for the delay i was having trouble with lerna locally.

how is this for a cli test? or do you think there's a better place or better test to do?

packages/cli/test/fixtures/build-modules/expected/unbundled-es/push-manifest.json
```diff
@@ -11,6 +11,13 @@
     "node_modules/dep1/dep1.js": {
       "type": "script",
       "weight": 1
+    },
+    "esm-mod1.mjs": {
+      "type": "script",
+      "weight": 1
+    },
+    "esm-mod2.mjs": {
+      "weight": 1
```
43081j: should this be here? if so, should it be missing a `type` like above has?

43081j:
CI failing at the min, not sure what went wrong. looks like integration tests failed but the change is in a polyfill? ill investigate

PR #680, opened 2018-08-31
Retry browser navigation in WCT if the browser doesn't connect.

bicknellr:
This retries 3 times, waiting at most 10s (or 1/2 of the total test timeout) each time. Posting this PR because Sauce Labs' Linux + Chrome 41 VM seems to start accepting WebDriver commands before the tunnel is up, so the single attempt to navigate hits their proxy's error page. If you go into their dashboard to control the test live, Chrome has already gone to the error page before you get there. You can refresh to actually start the tests because the proxy is up by the time you arrive.

- [x] tests


packages/web-component-tester/runner/browserrunner.ts
```diff
@@ -76,6 +92,13 @@ export class BrowserRunner {
     this.emitter = emitter;
     this.url = url;
 
+    this._navigationAttemptCount = 3;
+    this._navigationAttemptTimeout = Math.min(this.timeout / 2, 10000);
```
usergenic: I don't understand the `this.timeout / 2` bit.

bicknellr: This is meant to make sure that the time spent waiting before refreshing the page doesn't consume the total time allocated to the tests, even if the total timeout is small. It seemed like a good idea to me when writing this but I'm not actually convinced it's helpful now that you've pointed it out: it only comes into effect if `this.timeout < 20000`, which seems unreasonably short for a total timeout. Leaving it in would mean that there's always at least one retry attempt for any given run but the retry timeout scales down when the total timeout is small. Removing it would mean that retries only happen after an absolute 10s, even if doesn't leave enough time for the tests to run (if the total timeout is small).

I think this line is trying to handle an unlikely case with behavior that doesn't have a clear benefit. Given that, I'm leaning in favor of replacing it with just a constant 10s, but I'm up for whatever you want to do here.

usergenic: Cool!  Seems like a reasonable way to ensure the connection timeout leaves enough headroom for the rest of the test to execute prior to overall timeout.

usergenic:
Awesome PR, @bicknellr

bicknellr:
Thanks, glad to help!

PR #683, opened 2018-09-04
count for first failing launched test when we have a before hook fail…

myammt:
Tests that are run in a suite that has a before hook failure, are not detected as having failing status, because the test-start event is sent to browserrunner but not the test-end for the first test that run in this suite. In order to detect these failures any test that has unknown state (only start) is considered as failing test.


usergenic:
@myammt I like where this PR was going.  I was looking at implementing something similar recently because the 0 tests detected/run case is treated as "Tests passing" which is bad.  What was the reason you closed this, just curious?

PR #690, opened 2018-09-05
Faster HTML Stringify

### LINKED ISSUE #688: [Analyzer] HtmlDocument::stringify() is very slow
On a medium sized project, `HtmlDocument::stringify()` can take a great deal of time due to the use of `clone` to copy the entire HtmlDocument.

It seems like all that is needed is just a clone of the HTML ast, which is much smaller and faster to clone.

azakus:
- Clone only the HTMLDocument ast, and stringify inline documents into the
cloned ast at the same position.

Fixes #688

### DIFF
--- a/packages/analyzer/CHANGELOG.md
+++ b/packages/analyzer/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Make `HtmlDocument::stringify()` faster by only cloning the ast and stringifing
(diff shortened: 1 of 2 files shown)


packages/analyzer/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Make `HtmlDocument::stringify()` faster by only cloning the ast and stringifing
```
rictic: We've been using # to denote instance fields. So `HtmlDocument#stringify()`

azakus: Ok, will do

PR #709, opened 2018-09-17
Prepare release wct-local v2.1.2

usergenic:
- Changed postinstall.js script to accept a SELENIUM_OVERRIDES_CONFIG environment variable to allow specifying an alternate location for a json file containing a 'selenium-overrides' key with configuration details for the `selenium.install()` step.  Thanks to @bernardoVale for contribution.


packages/esm-amd-loader/tsconfig.json
```diff
@@ -7,8 +7,9 @@
     "preserveConstEnums": false,
     "sourceMap": false,
     "lib": [
+      "dom",
       "es5",
-      "dom"
+      "es6"
```
usergenic: So the updated `@types/node` acquired by `npm run nuke` has a reference to `Set` and even though node types are not technically used by esm-amd-loader, because they are hoisted, they are processed.  Without including `es6` in the tsconfig here, I don't know how to suppress the error encountered when running tsc.  Adding `../../node_modules/@types/node` to the ignore block doesn't seem to work.

justinfagnani: well, you shouldn't ever need both es5 and es6 at least.

usergenic: Ok. I'll flip it to be *just* es6 then for now.

usergenic:
```
+ wct-local@2.1.2
lerna success exec Executed command in 1 package: "npm publish"
```

PR #711, opened 2018-09-18
Fix resolution of absolute-path URLs (e.g. /foo.js).

### LINKED ISSUE #532: [esm-amd-loader] Not properly resolving files at root
### Description
When bundling and compiling using polymer build, if you use a root path (`/elements.js`) for the main shell, it is not properly resolved to the root.

The define function call shows `define(['/elements.js'])`, but on a url such as `/sub/page/`, the actual resolved request is `/sub/page//file.js`.

### Versions & Environment
- polymer-build: 1.7.3
- node: 9.11.2
- Operating System: Windows 10

### Steps to Reproduce

```json
{
 "entrypoint": "entrypoint.html",
 "shell": "elements.js",
 "extraDependencies": [
   "node_modules/@webcomponents/webcomponentsjs/*.js",
   "node_modules/@webcomponents/webcomponentsjs/bundles/*.js",
   "!node_modules/@webcomponents/webcomponentsjs/gulpfile.js"
 ],
 "builds": [
   {
     "name": "./",
     "preset": "es5-bundled",
     "addServiceWorker": false,
     "addPushManifest": false
   }
 ],
 "node": true,
 "moduleResolution": "node"
}
```
> entrypoint.html
```html
<!DOCTYPE html>
<html lang="en">
<head>
 <script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 <script type="module" src="/elements.js"></script>
</head>
<body>
  <paper-button>Button</paper-button>
</body>
</html>
```
> elements.js
```js
import '@polymer/paper-button/paper-button.js';
```

#### Steps
Build with the above file configuration, then try to serve entrypoint from a non-root directory.

#### Expected Results
Properly fetches `elements.js` and loads data.

#### Actual Results
```log
(index):38 Uncaught TypeError: Failed to fetch https://example.domain/sub/page//elements.js
   at HTMLScriptElement.f.onerror ((index):38)
```

aomarks:
Fixes #532

### DIFF
--- a/packages/esm-amd-loader/src/esm-amd-loader.ts
+++ b/packages/esm-amd-loader/src/esm-amd-loader.ts
@@ -545,7 +545,7 @@ function resolveUrl(urlBase: NormalizedUrl, url: string): NormalizedUrl {
     // Already a fully qualified URL.
     return url as NormalizedUrl;
   }
-  return normalizeUrl(urlBase + url);
+  return normalizeUrl(url[0] === '/' ? url : urlBase + url);
 }
 
 function getBaseUrl(): NormalizedUrl {
--- a/packages/esm-amd-loader/test/src/suite.ts
+++ b/packages/esm-amd-loader/test/src/suite.ts
@@ -91,6 +91,36 @@ suite('static dependencies', () => {
     });
   });
 
+  test('loads absolute path url', (done) => {
+    define(
+        ['/components/@polymer/esm-amd-loader-test/static/y/y.js'],
+        (y: any) => {
+          assert.equal(y.y, 'y');
+          done();
+        });
+  });
+
+  suite('with base tag', () => {
+    let base: HTMLBaseElement;
+
+    suiteSetup(() => {
+      base = document.createElement('base');
+      base.href = '/components/@polymer/esm-amd-loader-test/';
+      document.head.appendChild(base);
+    });
+
+    suiteTeardown(() => {
+      document.head.removeChild(base);
+    });
+
+    test('loads relative path url', (done) => {
+      define(['./static/y/y.js'], (y: any) => {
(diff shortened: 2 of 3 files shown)


packages/esm-amd-loader/src/esm-amd-loader.ts
```diff
@@ -545,7 +545,7 @@ function resolveUrl(urlBase: NormalizedUrl, url: string): NormalizedUrl {
     // Already a fully qualified URL.
     return url as NormalizedUrl;
   }
-  return normalizeUrl(urlBase + url);
+  return normalizeUrl(url[0] === '/' ? url : urlBase + url);
```
rictic: We should also document on normalizeUrl that it can take urls like `/foo/bar` as input

packages/esm-amd-loader/test/src/suite.ts
```diff
@@ -91,6 +91,36 @@ suite('static dependencies', () => {
...
+    suiteTeardown(() => {
+      document.head.removeChild(base);
+    });
+
+    test('loads relative path url', (done) => {
+      define(['./static/y/y.js'], (y: any) => {
+        assert.equal(y.y, 'y');
+        done();
+      });
+    });
```
rictic: Test absolute origin-level url with a base tag?

packages/esm-amd-loader/src/esm-amd-loader.ts
aomarks: Done.

packages/esm-amd-loader/test/src/suite.ts
aomarks: Done.

jsilvermist:
Awesome, thanks for the fix!

PR #715, opened 2018-09-18
[web-component-tester] runner always needs to inject mocha for the generated index

usergenic:
I forgot to inject mocha into the scripts to load and that's totally necessary or generated index or `wct-mocha/browser.js` will complain that Mocha isn't loaded.

Also, I removed the definition of a default `wctPackageName` option in `config.ts` since `webserver.ts` now assigns (if undefined) in terms of the first resolved WCT client package when attempting to resolve them in order of wct-mocha, wct-browser-legacy, web-component-tester.

### DIFF
--- a/packages/web-component-tester/runner/config.ts
+++ b/packages/web-component-tester/runner/config.ts
@@ -228,8 +228,9 @@ export function defaults(): Config {
       port: undefined,
       hostname: 'localhost',
     },
-    // The name of the NPM package that is vending wct's browser.js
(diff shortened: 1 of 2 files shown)


usergenic:
I caught the issue of Mocha not being injected in my wct-mocha integration tests (the only place it wasn't happening, since it happens for wct-browser-legacy and web-component-tester already).  I'm piece-mealing the mega PR over in chunks, so the coverage is coming in the add wct-mocha package PR.

Issue #568, opened 2018-07-09
[bundler] Multiple import() statements for the same fragment breaks bundles


keanulee:
When I edit:

https://github.com/Polymer/pwa-starter-kit/blob/b50a93748677c8be0bdcdc5ed71a03747628fc57/src/actions/app.js#L31-L39

```js
  switch(page) {
    case 'view1':
      import('../components/my-view1.js').then((module) => {
        // Put code in here that you want to run every time when
        // navigating to view1 after my-view1.js is loaded.
      });
      break;
    case 'view2':
      import('../components/my-view2.js');
      break;
```

To:

```javascript
    case 'view1':
      import('../components/my-view1.js').then((module) => {
        // Put code in here that you want to run every time when
        // navigating to view1 after my-view1.js is loaded.
      });

      import('../components/my-view2.js');

      break;
    case 'view2':
      import('../components/my-view2.js');
      break;
```

The resulting shell bundle (my-app.js) doesn't export some identifiers used by the fragments like `LitElement` and `html`.

Ref https://github.com/Polymer/pwa-starter-kit/issues/190

abdonrd:
@usergenic any news about this? 🙂

LarsDenBakker:
This is a serious bug, not very encouraging that it's been open since july :(

RobMaskell:
Strangely hit this two days ago and it took a while to unwind it

usergenic:
FWIW, working on this via Polymer/tools#722 - rollup has some unfortunate assumptions and its own new logic around deduplicating dynamic imports that I'm attempting to get around. Solution soon.

usergenic:
Filed another bug on rollup https://github.com/rollup/rollup/issues/2489

usergenic:
Fixed by https://github.com/Polymer/tools/pull/722

PR #722, opened 2018-09-26
[bundler] Fixed a couple important issues regarding our use of Rollup

usergenic:
- Fixed `external` identification by use of function in Rollup config, so we can reliably identify URLs for dynamic imports as external, so they are not inlined and/or deduplicated.
 - Resolve all module IDs before allowing Rollup to deal with them, because rollup's resolution doesn't properly deal with mixing relative with our resolved URLs.


keanulee:
Verified the fix works for multiple `import()` statements for the same fragment.

packages/bundler/src/es6-module-bundler.ts
```diff
@@ -78,7 +78,7 @@ async function prepareBundleModule(
         getOrSetBundleModuleExportName(assignedBundle, sourceUrl, '*');
     bundleSource.body.push(babel.importDeclaration(
         [babel.importNamespaceSpecifier(babel.identifier(starExportName))],
-        babel.stringLiteral(rebasedSourceUrl)));
+        babel.stringLiteral(sourceUrl)));
```
aomarks: Looks like `rebasedSourceUrl` isn't used anymore in this function? Remove?

packages/bundler/src/es6-rewriter.ts
```diff
@@ -68,11 +68,18 @@ export class Es6Rewriter {
     // resolutions for such things as bare module specifiers.
     const jsImportResolvedUrls = new Map<string, Map<string, ResolvedUrl>>();
     if (document) {
-      jsImportResolvedUrls.set(input, this._getJsImportResolutions(document));
+      jsImportResolvedUrls.set(input, this.getEs6ImportResolutions(document));
     }
     const rollupBundle = await rollup({
       input,
-      external,
+      external: (moduleId, parentId) => {
```
aomarks: Add comment like "external tells rollup not to bundle a particular import".

packages/bundler/src/es6-rewriter.ts
```diff
@@ -188,18 +202,50 @@ export class Es6Rewriter {
     return resolutions;
   }
 
+  rewriteEs6SourceUrlsToResolved(
+      node: babel.Node, jsImportResolvedUrls: Map<string, ResolvedUrl>) {
+    const bundler = this.bundler;
```
aomarks: unused variable?

packages/bundler/src/es6-rewriter.ts
```diff
@@ -188,18 +202,50 @@ export class Es6Rewriter {
...
+    const rewriteDeclarationSource = {
+      enter(
+          path:
+              NodePath<babel.ExportNamedDeclaration|babel.ImportDeclaration>) {
```
aomarks: Does this also need to handle ExportAllDeclarations?

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
```diff
@@ -313,5 +312,38 @@ suite('Es6 Module Bundling', () => {
         };
         export { a$1 as $a, b$1 as $b, a, b };`);
     });
+
+    test('deduplicate static imports, not dynamic', async () => {
```
aomarks: would this test case catch both of the issues mentioned?

packages/bundler/src/es6-module-bundler.ts
usergenic: Absolutely.  Good catch.  Removed.

packages/bundler/src/es6-rewriter.ts
usergenic: Comment'd!

packages/bundler/src/es6-rewriter.ts
usergenic: Totes unused.  Removed.  Thanks!

packages/bundler/src/es6-rewriter.ts
usergenic: Ugh. Rollup chokes on those:
```
export { b } from 'file:///memory/b.js';



var d = ({
        *file:///memory/b.js: *
});

export { d as $d };
```

Working on finding a version of Rollup that does the right thing.  Different versions choking in different ways... :(

usergenic:
Filed another bug on rollup which is kind of blocking the ExportAllDeclaration from working properly here.  https://github.com/rollup/rollup/issues/2489

usergenic:
Proposed solution is to transform `export * from` to `export {a, b, c} from` in the load step before I hand a file to rollup.  Analyzer should be able to provide the set of exported values in the `export *` case for me to do this.  This works around https://github.com/rollup/rollup/issues/2489 until rollup has a fix.

packages/bundler/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-<!-- ## Unreleased -->
+## Unreleased
+* Fix issue with multiple dynamic imports of the same fragment.
```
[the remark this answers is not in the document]
usergenic: Link added.

packages/bundler/src/es6-module-bundler.ts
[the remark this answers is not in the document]
usergenic: There should no longer be an issue of confusion here as `sourceUrl` is always a `resolvedUrl` since it comes from `for (const sourceUrl of [...assignedBundle.bundle.files].sort()) {`

The "rebasedSourceUrl" was a thing back when I was trying to keep them all relative URLs but now I only use resolveUrls until the final product is delivered out of the bundler.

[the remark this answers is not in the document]
usergenic: I could rename it to resolvedSourceUrl if that clarifies it.

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
usergenic: This test covers the issue #568 which boils down to not letting rollup dedup dynamic imports.  The other issues addressed in the PR are tested in https://github.com/Polymer/tools/pull/722/files#diff-11afbe94c33f83818dca10824afa21b6R107 for example.
Other issues tackled are not surfaced easily in tests; they're more about changing the use of rollup under-the-hood to enable the other things.  I think the existing tests demonstrate there is no regression.

Is that sufficient?

packages/bundler/src/es6-rewriter.ts
aomarks: Looks like this was resolved.

packages/bundler/src/test/es6-module-bundling-scenarios_test.ts
aomarks: ok

abdonrd:
@usergenic could you release a new `polymer-cli` version with this new version?

Westbrook:
@usergenic This code is in polymer-bundle@4.0.3, right? I'm still having issue building with lit-html and not running into import confusion, was there something more than just an update required to get this update working?

Issue #726, opened 2018-09-27
[wct] Tests run multiple times when using dynamic modules polyfilled with AMD


bashmish:
## Description

Polymer CLI (e.g. `polymer serve`) injects code for WCT to defer execution of tests until modules are loaded. This injected code breaks execution of tests if the code contains dynamic modules.

Example of broken tests:

```javascript
import myStaticModule from './my-static-module.js';

let beforeEachCount = 0;
let test1Count = 0;
let test2Count = 0;

describe('run twice', () => {
  beforeEach(() => {
    beforeEachCount += 1;
    console.log('beforeEachCount:', beforeEachCount);
  });

  it('test 1', async () => {
    await import('./my-dynamic-module.js');
    test1Count += 1;
    console.log('test1Count:', test1Count);
    expect(test1Count, 'test 1 run twice').to.not.eql(2);
  });

  it('test 2', async () => {
    await import('./my-dynamic-module.js');
    test2Count += 1;
    console.log('test2Count:', test2Count);
    expect(test2Count, 'test 2 run twice').to.not.eql(2);
  });
});
```

Injected code by Polymer CLI which breaks the execution:

```html
<script>
  // Injected by polymer-build to defer WCT until all AMD modules are loaded.
  (function() {
    var originalDefine = window.define;
    var moduleCount = 0;
    window.define = function(deps, factory) {
      moduleCount++;
      originalDefine(deps, function() {
        if (factory) {
          factory.apply(undefined, arguments);
        }
        moduleCount--;
        if (moduleCount === 0) {
          window._wctCallback();
        }
      });
    };
  })();
</script>
```

## Expected Result

As soon as `moduleCount` reaches `0` the callback `window._wctCallback();`is run. After that this callback is not run anymore.

## Actual Result

All static modules work as expected incrementing the `moduleCount` and after that it's value is decremented until `0` and the callback is run for the first time. After that dynamic modules call `define` asynchronously causing the increment of `moduleCount` again and then the decrement leading to running the callback for the second time. Then the callback runs Mocha under the hood for the second time (check `_runMocha` function in `web-component-tester/browser.js`) leading to tests running for the second time. Tests fail because they have been already finished before messing up the output like if they are new ones.

## Browsers Affected

- Firefox
- Edge
- IE
- ... any missing native support for both static and dynamic imports

## Demo

I created a repo where you can easily see this in action: https://github.com/bashmish/bug-polymer-serve-wct-dynamic-import-run-twice

## Screenshot

[...]

PR #727, opened 2018-09-27
Fix injected WCT hack to prevent dynamic modules run Mocha multiple times

bashmish:
Fixes https://github.com/Polymer/tools/issues/726

As far as I understand the intention of the injected code it should be safe to assume that all static modules will be loaded by the time the callback is run for the first time. This is why I just disable it for future calls so that dynamic imports do not cause the issues.

I tested this fix on our code containing multiple suites all using components which use dynamic imports. Works well.

### DIFF
--- a/packages/build/src/html-transform.ts
+++ b/packages/build/src/html-transform.ts
@@ -284,6 +284,7 @@ function addWctTimingHack(wctScript: dom5.Node, amdLoaderScript: dom5.Node) {
         moduleCount--;
         if (moduleCount === 0) {
           window._wctCallback();
+          window.define = originalDefine;
         }
       });
     };
--- a/packages/polyserve/test/bower_components/test-modules/golden/test-suite-wct.html
+++ b/packages/polyserve/test/bower_components/test-modules/golden/test-suite-wct.html
@@ -37,6 +37,7 @@
         moduleCount--;
         if (moduleCount === 0) {
           window._wctCallback();
+          window.define = originalDefine;
         }
       });
     };


bashmish:
@TimvdLippe I would love to. So far I found that you do not test the impact of the injected code and only test that it was injected at a proper place in the page.

https://github.com/Polymer/tools/blob/dc391830fec747bb6164109157cd3a24302e346a/packages/build/src/test/html-transform_test.ts#L44

https://github.com/Polymer/tools/blob/dc391830fec747bb6164109157cd3a24302e346a/packages/build/src/test/html-transform_test.ts#L365-L376

Can you please point me at the tests which directly or indirectly test the behavior of the injected code?

TimvdLippe:
Hm, I think @aomarks knows more about that. Al, do you mind helping out with testing these features?

bashmish:
@TimvdLippe do I understand correctly that the integration tests currently fail because of some selenium/saucelabs issue and not because of a test error here https://travis-ci.org/Polymer/tools/builds/434066022?utm_source=github_status&utm_medium=notification

TimvdLippe:
@bashmish Sadly these issues are due to Sauce flakiness. However, they are absolutely unrelated to your work :smile:

dlockhart:
@bashmish @TimvdLippe Any update on this fix? I'm running into this issue as well with a LitElement that does a dynamic import.

TimvdLippe:
This would still require a review by @aomarks

dlockhart:
@bashmish any chance this could get a review? We're still getting these errors when we use dynamic imports and it's forcing us to disable our tests in Edge and IE11... which isn't ideal. Thanks!

aomarks:
The sauce flake is fixed. A test would be nice, but I don't think it's necessary here. I'll add a CHANGELOG separately, and this will go out with the next WCT release.

bashmish:
Thanks for finally merging this!

However, for everyone who wants to switch to more up-to-date solution for testing web components: I encourage you to visit this https://github.com/open-wc/open-wc/tree/master/packages/testing-karma

Generally this requires you to adapt your tests a little bit, practically meaning you need to move tests to JS files, provide a few globals like `flush` yourself and use new helpers from https://github.com/open-wc/open-wc/tree/master/packages/testing-helpers for defining fixtures (or even replace `flush` with `aTimeout` or smth like this which is most of the time OK).

svanherk:
Hi @aomarks, do you know when the next WCT release will happen?  We're eagerly awaiting this change! 🎉

svanherk:
@TimvdLippe @aomarks @justinfagnani Checking in on this again, will there be any further releases of WCT?  We are slowly moving over to `open-wc`'s testing tools, but we have a number of old repos still using `wct` and I'm hoping we can re-enable Edge tests soon.

PR #729, opened 2018-09-27
Turn off deadcode removal

### LINKED ISSUE #724: lit-html render function breaks when minifying and compiling to es5
Hi,

I am not sure if this is the right place to report this issue, but since it is a problem with the code compilation, it seems wise.

You can see below the two versions of the compiled code for the `render` function of the `lit-html` package when using the Polymer LitElement.

ES5 (not minified)
```javascript
function render(result, container) {
    var templateFactory$$1 = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : templateFactory;
    var part = parts$1.get(container);

    if (part === undefined) {
      removeNodes(container, container.firstChild);
      parts$1.set(container, part = new NodePart(templateFactory$$1));
      part.appendInto(container);
    }

    part.setValue(result);
    part.commit();
}
```

ES5 (minified)
```javascript
function render(result, container) {
    var templateFactory$$1 = 2 < arguments.length && arguments[2] !== void 0 ? arguments[2] : templateFactory,
        part = parts$1.get(container);
    part.setValue(result);
    part.commit();
}
```

As you can see the check of the `part` is missing which results to the following error `TypeError: Cannot read property 'setValue' of undefined at render`.

These are the contents of the `polymer.json` file I am using.
```
{
  "npm": true,
  "lint": {
    "rules": [
      "polymer-3"
    ]
  },
  "entrypoint": "src/index.html",
  "shell": "src/main.js",
  "moduleResolution": "node",
  "extraDependencies": [
    "node_modules/@webcomponents/webcomponentsjs/**"
  ],
  "builds": [
    {
      "name": "esm-bundled",
      "browserCapabilities": [
        "es2015",
        "modules"
      ],
      "js": {
        "minify": true
      },
      "css": {
        "minify": true
      },
      "html": {
        "minify": true
      },
      "bundle": true
    },
    {
      "name": "es2015-bundled",
      "browserCapabilities": [
        "es2015"
      ],
      "js": {
        "compile": "es2015",
        "minify": true,
        "transformModulesToAmd": true
      },
      "css": {
        "minify": true
      },
      "html": {
        "minify": true
      },
      "bundle": true
    },
    {
      "name": "es5-bundled",
      "js": {
        "compile": "es5",
        "minify": true, // it works if I set this to false
        "transformModulesToAmd": true
      },
      "css": {
        "minify": true
      },
      "html": {
        "minify": true
      },
      "bundle": true
    }
  ]
}
```

Any ideas?

Westbrook:
Fixes #724

As per https://github.com/babel/minify/issues/574#issuecomment-307390713 there are times when actual code can be removed as if it were dead code.

I'm working on a better solution, but wanted to get a better run on the tests here than I'm having luck for locally.

### DIFF
--- a/packages/build/src/js-transform.ts
+++ b/packages/build/src/js-transform.ts
@@ -97,23 +97,29 @@ const babelSyntaxPlugins = [
   require('@babel/plugin-syntax-import-meta'),
 ];
 
-const babelPresetMinify = require('babel-preset-minify')({}, {
-  // Disable the minify-constant-folding plugin because it has a bug relating to
-  // invalid substitution of constant values into export specifiers:
-  // https://github.com/babel/minify/issues/820
-  evaluate: false,
+// Process as a method so that the options can be used to determine
+// whether to eliminate dead code or not. As per
+// https://github.com/Polymer/tools/issues/724 compiling to `es5` when
+// removing dead code can cause live code to be erroneously removed.
+function babelPresetMinify(options: JsTransformOptions) {
+  return require('babel-preset-minify')({}, {
+    // Disable the minify-constant-folding plugin because it has a bug relating to
+    // invalid substitution of constant values into export specifiers:
+    // https://github.com/babel/minify/issues/820
+    evaluate: false,
 
-  // TODO(aomarks) Find out why we disabled this plugin.
-  simplifyComparisons: false,
+    // TODO(aomarks) Find out why we disabled this plugin.
+    simplifyComparisons: false,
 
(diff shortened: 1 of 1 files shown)


Westbrook:
I'd really like to get the tests passing for this, but am unable to comprehend the report from https://travis-ci.org/Polymer/tools/jobs/434586205 Would anyone be able to support me in understanding what it's telling me?

The only actionable part I see is `Chrome version must be >= 67.0.3396.0\n` but there is a lot more report after that... 😖

PR #677, opened 2018-08-29
all: standardise tslint config

43081j:
Fixes #305.

---

Let me know if you still want this, otherwise i will throw it away.

* Added missing lint commands
* Made all tslint configs inherit the base one
* Fixed a couple of lint errors
* Added a lint script (`npm run lint`) to lint all packages

Couple of packages break the `no-any` rule so i turned it off as i didn't want to be altering interfaces for this.

PR #733, opened 2018-10-04
Added lint to packages missing it and made linting a final build step.

usergenic:
- I wanted to start ensuring lint passes are running before we commit things and it felt like earlier in the workflow, i.e. build, was a better phase than test.
- Added lint to all typescript packages
- Added `curly: true` and fixed code where lint reported typings were too vague.
- I couldn't get all packages on the top-level lint config yet because, especially with dom5, so much hassle.  But this gets us moving towards that.


TimvdLippe:
I think this is also being done in #677 ?

packages/workspaces/tslint.json
```diff
@@ -1,60 +1,61 @@
 {
```
rictic: It looks like this has both /tslint.json and /packages/workspaces/tslint.json?

usergenic:
Merged #677 and resolved conflicts in this branch.

packages/workspaces/tslint.json
usergenic: Resolved: workspaces/tslint.json is now just extending /tslint.json

rictic: Why have both at all?

package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "lerna run lint --stream",
     "build": "lerna run build --stream",
     "format": "lerna run format --stream",
+    "lint": "lerna run lint --stream",
```
43081j: This can be removed i think as it is duplicated above from the merge of the other lint PR

usergenic: Thanks! Removed.

packages/workspaces/tslint.json
aomarks: ping?

PR #734, opened 2018-10-09
Rerun flaky integration tests

usergenic:
I am so sick of (WCT in particular) integration tests on tools monorepo blowing an entire CI run because of one or two hiccups.  My solution was to write a custom mocha reporter that spawns another process to execute the same mocha script, but including a grep for the failed test titles.

Right now I've set up both `web-component-tester` and `polymer-cli` to use this reporter.  I haven't made it its own package yet, but I may do that next since its kind of weird to have `polymer-cli` reach into WCT's folder to get at it.

Note: I spawn a new process because mocha is bad at rerunning tests in the same process due to the way it uses require() to load Mocha.run() and the require.cache foils attempts to rerun.  Blowing the require() cache is actually not 100% reliable and since a failure to do so results in invisibly not finding the tests in question, I felt a more reliable approach was needed.


packages/web-component-tester/test/retry-failures-reporter.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Google as part of the polymer project is also subject to an additional IP
+ * rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import * as chalk from 'chalk';
+import {spawnSync} from 'child_process';
+import * as Mocha from 'mocha';
+
+const retryTargets: string[] =
+    JSON.parse(process.env['TEST_RETRY_TARGETS'] || '[]');
```
justinfagnani: JSON.parse can throw

packages/web-component-tester/test/retry-failures-reporter.ts
```diff
@@ -0,0 +1,161 @@
...
+const retryTargetMax =
+    parseInt(process.env['TEST_RETRY_TARGET_MAX'] + '', 10) || 0;
```
justinfagnani: why the `+ ''`?

usergenic: Oh derp.  I didn't mean to leave that in.  I was worried an undefined value would throw but parseInt handles undefined just fine.  Removed the unnecessary `+ ''` concatenation.

PR #737, opened 2018-10-15
Standardized tslint.json for wct-local

usergenic:
Another small step towards final standardization of tslint.json in all packages


usergenic:
Thx, @aomarks

PR #744, opened 2018-10-18
Updated bundler CHANGELOG and patch version number for release.

usergenic:
This is a patch release to fix the broken polymer-bundler CLI `-h` option (was due to breaking change in chalk brought in by command-line-usage package.


usergenic:
Thx @aomarks

PR #745, opened 2018-10-19
Updated version and CHANGELOG for prerelease of wct-mocha

usergenic:
A version of wct-mocha 1.0.0-pre.1 was accidentally published as "latest" (and then unpublished) a month or so ago, which consumed the 1.0.0-pre.1 version, so this will be published as 1.0.0-pre.2


usergenic:
Thx @aomarks

PR #749, opened 2018-10-23
Firefox push fixed in 63

keanulee:
Verified with https://20170915t110514-dot-polymer-shop.appspot.com/, the last version of Shop pushed with an older version of browser-capabilities that pushed to Firefox. Ref https://bugzilla.mozilla.org/show_bug.cgi?id=1409570


keanulee:
Integration tests are failing (ironically) on Firefox 63 - something with selenium?

usergenic:
@keanulee yeah the coincidence of FF 63 gave me pause.  Maybe we can pin FF to 62 in the travis config...  Looking into it.

usergenic:
I'm not actually clear on the best way to pin FF version on Travis and/or if its possible to set it to version 62.  I'm going to test locally with FF63 and see if there's any general Selenium + FF63 issues

Issue #775, opened 2018-10-31
polymer-bundler v3.1.1 fails to inline script with empty getter function


freshp86:
bug.html file

```html
<script src="bug.js"></script>
```

bug.js file
```js
class Foo {
  get foo() {}
}
```

Polymer bundler command (version 3.1.1)
```
./node_modules/polymer-bundler/lib/bin/polymer-bundler.js --in-html bug.html --inline-scripts --manifest-out manifest.json
```

The command above returns the contents of the original HTML file without inlining the JS file. The contents of the manifest indicate that something went wrong.
```json
{"bug.html":["bug.html"],"_missing":["bug.js"]}
```

Apparently polymer-bundler failed to find the JS file, which seems very odd.

If `get foo() {}` is changed to have some contents, like 

```js
get foo {
  console.log('foo');
}
```
The problem no longer happens, and the script is inlined as expected.

moderndeveloperllc:
@usergenic Issue with dead code removal? Just a thought. Seen lots of those floating around babel repos lately for some reason.

usergenic:
@moderndeveloperllc makes a lot of sense assuming the dead-code removal occurs prior to bundling.  Verifying...

PR #777, opened 2018-11-01
Disable deadcode removal in babel-minify

usergenic:
Fix issue https://github.com/Polymer/tools/issues/724 where deadcode removal in babel-minify can be error prone and result in removal of code which should not be removed and cause hard to debug errors.

### DIFF
--- a/packages/build/src/js-transform.ts
+++ b/packages/build/src/js-transform.ts
@@ -99,8 +99,7 @@ const babelSyntaxPlugins = [
 
 const babelPresetMinify = require('babel-preset-minify')({}, {
   // Disable the minify-constant-folding plugin because it has a bug relating
-  // to
-  // invalid substitution of constant values into export specifiers:
+  // to invalid substitution of constant values into export specifiers:
   // https://github.com/babel/minify/issues/820
   evaluate: false,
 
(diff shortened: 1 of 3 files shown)


usergenic:
There are 3 open issues with deadcode removal that may or may not be directly contributing here, but more importantly, until we can fully rely on deadcode removal behavior, I think we should disable this optimization.

https://github.com/babel/minify/issues/879
https://github.com/babel/minify/issues/895
https://github.com/babel/minify/issues/898

packages/build/src/js-transform.ts
```diff
@@ -98,14 +98,20 @@ const babelSyntaxPlugins = [
 ];
 
 const babelPresetMinify = require('babel-preset-minify')({}, {
-  // Disable the minify-constant-folding plugin because it has a bug relating to
+  // Disable the minify-constant-folding plugin because it has a bug relating
+  // to
```
rictic: reflow?

usergenic: derp! thanks!

moderndeveloperllc:
@usergenic Does this also solve #775?

usergenic:
@moderndeveloperllc I was able to replicate that problem with current version of bundler.  Solving for the 3.x branch may be possible but I've not yet identified what's causing it.

usergenic:
(edited: "I was _not_ able to replicate")

PR #783, opened 2018-11-10
Updated package locks and goldens in preparation for cli and build releases.

usergenic:
Putting these changes in their own PR as the updated goldens were muddying the waters on the actual release details.

Additionally, for as-yet-unknown-reasons, there is one integration test involving the es5-bundled shop-cart.js in the Polymer/shop lit-element branch which fails ONLY on headless chrome.  That was a real PITA to figure out.

As such I've turned off headless for that particular suite of integration tests.

Also, I discovered that the retry reporter (only on retry attempt) when loaded from the mocha for CLI from web-component-tester package folder would not be found because of path changing in mocha so I explicitly use `$(pwd)` in the bash script run by npm script to make it an absolute path, which avoids mocha getting confused about where to resolve it from.  So that was an opportunistic fix unrelated to the package-locks and goldens.


packages/web-component-tester/test/retry-failures-reporter.ts
```diff
@@ -78,7 +80,7 @@ module.exports = class RetryFailuresReporter extends Mocha.reporters.Spec {
           fails.map((t) => ' - ' + t).join('\n')));
 
       const retryResults = spawnSync(process.argv0, args, {
-        cwd: process.cwd(),
+        cwd: initialCurrentWorkingDirectory,
```
usergenic: Ensures we use the cwd() at the beginning of the run instead of whatever it might be when we are done.

Issue #736, opened 2018-10-15
[build] polymer build does not handle .mjs files


keanulee:
Unlike `polymer serve`, `polymer build` doesn't know to treat .mjs files as JavaScript files.

```
import {
  createStore,
  compose,
  applyMiddleware,
  combineReducers
} from 'redux/es/redux.mjs';
       ~~~~~~~~~~~~~~~~~~~~

file:///Users/keanulee/Code/Polymer/pwa-starter-kit/src/store.js(16,8) error [could-not-load] - Unable to load import: No parser for for file type mjs
```

Another example: https://github.com/Polymer/pwa-starter-kit/pull/249

aomarks:
Why not just use .js files?

See also https://github.com/Polymer/tools/issues/671

keanulee:
This would be more applicable to 3p libraries, for example https://unpkg.com/redux@4.0.1/es/redux.mjs (https://github.com/reduxjs/redux/pull/3143 from @TimvdLippe ).

rjsteinert:
I've opted for an awkward install step to fix the polymer builds in my project.

```
npm install && mv node_modules/redux/es/redux.mjs node_modules/redux/es/redux.js
```

mathiasbynens:
Patch: #792. Regardless of any one project’s decision to use `*.mjs` or `*.js` for its source files, tooling _must_ support `*.mjs` simply because people are using it.

bicknellr:
IMO, #792 seems reasonable given the current "file extension implies media type" behavior of the analyzer. It might be nice to split this process into two steps at some point to help out with forward compatibility:

Rather than just
```typescript
parsers = new Map<string, Parser<ParsedDocument>>([
  ['html', new HtmlParser()],
  ['js', new JavaScriptParser()],
  ['mjs', new JavaScriptParser()],
  ['css', new CssParser()],
  ['json', new JsonParser()],
]);
```
maybe doing something more like
```typescript
defaultMediaType = new Map<string, string>([
  ['html', 'text/html'],
  ['css', 'text/css'],
  ['js', 'application/javascript'],
  ['mjs', 'application/javascript'],
  ['json', 'application/json'],
  ...userSuppliedDefaultMediaTypes
]);

parsers = new Map<string, Parser<ParsedDocument>>([
  ['text/html', new HtmlParser()],
  ['text/css', new CssParser()],
  ['application/javascript', new JavaScriptParser()],
  ['application/json', new JsonParser()],
]);
```
[...]

PR #792, opened 2018-11-23
Recognize .mjs as JavaScript files

mathiasbynens:
This is a quick and dirty fix for https://github.com/Polymer/tools/issues/736.

### DIFF
--- a/packages/analyzer/src/core/analysis-context.ts
+++ b/packages/analyzer/src/core/analysis-context.ts
@@ -71,6 +71,7 @@ export class AnalysisContext {
   readonly parsers = new Map<string, Parser<ParsedDocument>>([
     ['html', new HtmlParser()],
     ['js', new JavaScriptParser()],
+    ['mjs', new JavaScriptParser()],
     ['css', new CssParser()],
     ['json', new JsonParser()],
   ]);


packages/analyzer/src/core/analysis-context.ts
```diff
@@ -71,6 +71,7 @@ export class AnalysisContext {
   readonly parsers = new Map<string, Parser<ParsedDocument>>([
     ['html', new HtmlParser()],
     ['js', new JavaScriptParser()],
+    ['mjs', new JavaScriptParser()],
```
mathiasbynens: Of course, you could create a single `const jsParser = new JavaScriptParser();` and then use the the same instance for both `js` and `mjs`.

mathiasbynens: You can also skip all the `import`/`export` scanning for `.mjs` files.

mathiasbynens:
What's blocking this? @bicknellr @justinfagnani @rictic @aomarks @usergenic

packages/analyzer/src/core/analysis-context.ts
rictic: Yeah, we want to add more contextual inference for treating a file as a script vs module, and .mjs would be a strong signal.

No objection to landing this as is though.

PR #799, opened 2018-12-01
Allow mixin classes to have members added via .prototype calls.

aomarks:
For example:

```js
  /** @type {string} */
  MyMixin.prototype.foo;
```

We already support this behavior for normal classes, but not for mixin classes.

We need this because we're going to change some class mixin properties such as `PolymerElement._template` to not be initialized in the constructor, but we still need a way to tell analyzer that they exist.

cc @sorvell @kevinpschaaf


rictic:
We can also do this via:

```javascript
/** @type {string} */
this.foo;
```

in  the constructor, which I'd argue is more legible. I guess if that's the only thing in the constructor it might be better for the VM to leave it off though?

aomarks:
> We can also do this via:
> 
> ```js
> /** @type {string} */
> this.foo;
> ```
> 
> in the constructor, which I'd argue is more legible. I guess if that's the only thing in the constructor it might be better for the VM to leave it off though?

Yeah @kevinpschaaf and @sorvell specifically wanted to avoid having the property accesses at all in the constructor, for uncompiled users.

PR #800, opened 2018-12-03
Prepare analyzer 3.2.0 release.

aomarks:
- [ ] polymer-analyzer@3.2.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,14 +4,14 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@lerna/add": {
-      "version": "3.4.1",
-      "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.4.1.tgz",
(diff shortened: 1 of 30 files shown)


usergenic:
(approved except for builds breaking...)

usergenic:
Assume this is because `npm run nuke` updated TypeScript to 3.2.1 from 3.1.6?

aomarks:
> Fix typings or pin TS back to 3.1.6?

I can fix the typings.

aomarks:
> I can fix the typings.

I think I fixed the typings. Not very proud of the fixes, they seem worse in every case.

usergenic:
I don't think the changes are all that bad.

The ones in the tests are to be expected because of the way `.args` is going to be ambiguous.  I do agree that adding `as any` into the code at places feels bad, but I think that's often more a matter of some limitations in the typings we're getting for libraries like with sinon etc.

Anyways lgtm

PR #805, opened 2018-12-06
Prepare esm-amd-loader 1.0.4 and cli 1.9.2 releases

keanulee:
- [ ] @polymer/esm-amd-loader@1.0.4
- [ ] polymer-cli@1.9.2

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -9,10 +9,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- Add new, unreleased changes here. -->
 
 ## v1.9.2 [12-06-2018]
-* Updated dependecies.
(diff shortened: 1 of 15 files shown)


packages/cli/CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ## Unreleased -->
 <!-- Add new, unreleased changes here. -->
 
+## v1.9.2 [12-06-2018]
+* Updated dependecies.
```
aomarks: "dependencies"

but could you actually summarize the relevant changelog items here since otherwise there's no information about what changed

keanulee: Updated. PTAL

PR #810, opened 2018-12-14
Fix typings in editor-service tests to make tools build again

keanulee:
This seems to make builds go again after `npm run nuke` at least. Let's see if the tests pass...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,8 @@ install:
   - npm ci
   - npm run bootstrap
   - npm run build
+  # NOTE(keanulee): Following needed until Chrome in AppVeyor images are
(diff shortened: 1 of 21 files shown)


keanulee:
Can we retry tests on appveyor? Seems like an unrelated issue.

keanulee:
Updated appveyor.yml with suggested workaround from https://github.com/appveyor/ci/issues/2772. PTAL and merge @usergenic @aomarks

appveyor.yml
```diff
@@ -11,6 +11,7 @@ install:
   - npm ci
   - npm run bootstrap
   - npm run build
+  - choco install googlechrome
```
aomarks: can you add a comment about why we need this?

PR #809, opened 2018-12-13
Update WCT dependencies to reduce vulnerabilities in npm security audit

usergenic:
- @types/gulp ^3.8.8 -> ^4.0.0
 - grunt ^0.4.5 -> ^1.0.3
 - gulp ^3.8.8 -> ^4.0.0
 - gulp-bower 0.0.13 -> 0.0.14
 - send ^0.11.1 -> ^0.16.1

Remaining WCT audit issues:
```
=== npm audit security report ===                        
                                                                                
# Run  npm install lodash@4.17.11  to resolve 1 vulnerability
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/577                       │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌──────────────────────────────────────────────────────────────────────────────┐
│                                Manual Review                                 │
│            Some vulnerabilities require your attention to resolve            │
│                                                                              │
│         Visit https://go.npm.me/audit-guide for additional guidance          │
└──────────────────────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=4.17.5                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ stacky                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ stacky > lodash                                              │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/577                       │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Prototype Pollution                                          │
[...]

PR #811, opened 2018-12-14
Audit deps

keanulee:
Built on top of #810 

- Updates cli to use `"github": "^13.0.0"` to remove `https-proxy-agent`. This is not the latest version (now https://www.npmjs.com/package/@octokit/rest), but is the first version that passes the audit without breaking us.
- Updates `send` and `lodash` in web-component-tester. I saw there's also #809, which seems like a superset of these changes (though introduces more source changes)

This doesn't fix all audit warnings, but should cover most of the more pressing ones.


keanulee:
Oh no. Don't merge this, but do take a look at #810

PR #812, opened 2018-12-15
[modulizer] Add new special case to convert Polymer.Element to PolymerElement

aomarks:
In the commit below, the Polymer base class declaration was simplified and is now a call expression instead of an identifier. We had a special case hack to do the Polymer.Element => PolymerElement rename, but it only worked in the identifier case.
    https://github.com/Polymer/tools/blob/ce5ca6cf050e2f72b41cd053c2e59f66bdb0c747/packages/modulizer/src/passes/rewrite-namespace-exports.ts#L249

Fixes https://github.com/Polymer/polymer/issues/5447

Note to reviewer: The first commit updates to the latest Polymer 2.x fixture, which shows the regression. The second commit fixes the regression.

cc @web-padawan


lkraav:
Just curious, a metric ton of the 729 files diff changes seems to be formatting and other stuff, was that intentional? It doesn't seem to be mentioned in any commit title or message, etc.

TimvdLippe:
These files are fixtures. We check these in the repository so that we can track the generated output. This way, it is very easy to spot potential regressions, as all the fixtures would be changed.

PR #817, opened 2018-12-15
Use the "resolved" exclude path

rcashie:
Not sure if this was intentional or I'm just using it wrong but it seems as though paths / dirs specified via the excludes param must be absolute.

The following small change uses the resolved path so that relative paths work.


usergenic:
Thanks, @rcashie - This looks great.  Merging.

usergenic:
Turns out I just started resolving excludes on Bundler construction so this change is obviated by that change but you still win the merge commit.  Thanks for submitting this.

PR #824, opened 2018-12-20
Update npm-ignore rules for all npm published packages

usergenic:
- Normalized all the `.npmignore` files to standard form with leading `/` and removed unnecessary automatic exclusions like `node_modules` and missing files like `.clang-format`.
 - Updated CHANGELOGs for packages which had `.npmignore` changes so the next release picks them up.
 - Opportunistic clean-up here in that I also removed unused per-package-top-level files like `.github` and `appveyor.yml`.
 - One or two `.js` files may have had formatting slightly updated because clang-format.

Here's a gist of the results of the new ignores: https://gist.github.com/usergenic/04c04c01623c0c8342bdbfb336cfdf63


usergenic:
@rictic you make a good point.  Probably best way to visualize the effects of changes and mitigate risk is to do packages as separate PRs including a diff of npm packaged files in the descriptions.

rictic:
Yeah, that'd be way easier to review! Sorry for the added friction!

PR #835, opened 2018-12-27
Add objectWithoutPropertiesLoose to generate-helpers.js

RobMaskell:
The babel helper objectWithoutProperties calls objectWithoutPropertiesLoose which needs to be included in generate-helpers.js in the build package in order to avoid an error. 
I found this while testing with "@babel/core": "^7.2.2" which fixes an issue I had with rest/spread operators not working in the version of babel used by polymer.


RobMaskell:
I signed it!

RobMaskell:
Doesn't look like the build failure is anything to do with my change, seems to be expecting Chrome 69+ but getting 62

RobMaskell:
Might be related to this https://github.com/Polymer/tools/commit/2f10194ada0f1f58447c2f5c5e94ccf92bf02671

RobMaskell:
@aomarks does this need to be released on it's own of does it get sucked by something else?

armujahid:
It's strange but I am still getting this error in IE from code compiled using latest polymer cli v1.9.11
`Object doesn't support property or method 'objectWithoutPropertiesLoose'`
I can see that generate helper file has `objectWithoutPropertiesLoose` so I am not sure what's the root cause.
https://github.com/Polymer/tools/blob/1fe1c22bfb81ad9195c99fa31ac24c675d15700c/packages/build/scripts/generate-helpers.js#L76

Issue #3323, opened 2019-01-10
[polymer-bundler] Inlining external module script fails when file has no export/import statements


usergenic:
```html
<!-- example.html -->
<script src="./module.js" type="module"></script>
```
```js
// module.js
console.log('hello');
```
```shell
$ polymer-bundler --inline-scripts example.html
<script src="./module.js" type="module></script>
```
then add an export to module.js...
```js
// module.js
export const pony = '🐴';
console.log('hello');
```
```shell
$ polymer-bundler --inline-scripts example.html
<script type="module>
const pony = '🐴';
console.log('hello');
</script>
```

This has to do with the way we determine whether the javascript document is a module or not through a contextless does this analyze-as-a-script-then-its-a-script-otherwise-analyze-as-a-module thing.  We should instead key off the type in the script tag.

usergenic:
Fixed and published polymer-bundler@4.0.5

PR #3326, opened 2019-01-10
Fixed issue #3323 where external module script tag content not inlined

usergenic:
Old logic tried to check Analyzer's assessment of the imported JavaScriptDocument to find out if it was a module, but that assessment is based on the presence of import/export statements instead of the script tag's `type` attribute.

This fix uses the ScriptTagImport instance's `isModule` attribute now, which is consistent regardless of the JavaScriptDocument properties.


usergenic:
cc @freshp86

usergenic:
Thx @aomarks

PR #3335, opened 2019-01-24
Additional hostname and port for accessing tests from an external network

marcojonker:
When running selenium from docker (for example zalenium), selenium calls the url for accessing the tests from a docker network. To access a webserver on the host machine from docker the hostname in the url needs to be 'host.docker.internal'. Currently the url is build with the same hostname as specifiek in webserver.hostname (default 127.0.01). I could change this to the IP address of the machine but using 'host.docker.internal' is much more generic. And since the webserver does not run with 'host.docker.internal' as a hostname an additional property is added to specify an additional hostname and port for accessing the tests.

To have more flexibility in formatting the url called for tests the properties 'testsHostname' and 'testsPort' are added to the webserver. These properties can be used to override 'hostname' and 'port' when formatting the url for running tests.

The might also be helpful for other situations when calling the tests from a remote network for example or through a proxy.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,12 +4,12 @@
   "lockfileVersion": 1,
   "dependencies": {
     "@lerna/add": {
-      "version": "3.10.2",
-      "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.10.2.tgz",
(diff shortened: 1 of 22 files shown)


marcojonker:
I signed it!

Issue #3353, opened 2019-02-11
[travis] Builds failing due to Selenium Chrome Driver


alfonso-presa:
Somehow Chrome Driver is failing when trying to start (travis is providing chrome 62 and driver is only compatible with >69). See [stack trace here](https://travis-ci.org/Polymer/tools/jobs/491633759#L684):

```
@polymer/esm-amd-loader: chrome 62                Tests failed: {"value":
{"stacktrace":"org.openqa.selenium.SessionNotCreatedException: session not created: Chrome version must be >= 69.0.3497.0\n  (Driver info: chromedriver=2.43.600233 
(523efee95e3d68b8719b3a1c83051aa63aa6b10d),platform=Linux 4.4.0-101-generic x86_64) 
(WARNING: The server did not provide any stacktrace information)\nCommand duration or timeout: 60.15 seconds\nBuild info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'\nSystem info: host: 'travis-job-1d5a7da7-f06e-4ac0-b8dc-1014bac8a99d', ip: '127.0.1.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.4.0-101-generic', java.version: 
'1.8.0_151'\nDriver info: driver.version: unknown\n\tat sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)\n\tat sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)\n\tat sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)\n\tat java.lang.reflect.Constructor.newInstance(Constructor.java:423)\n\tat 
```
-------
Edit: old content
I made a [PR](https://github.com/Polymer/tools/pull/3352) and the build failed. Digging a little bit inside the issue I noticed Saucelabs is being used and, very wisely, it's credentials are encrypted, which makes travis make them not available during  [unsafe PRs builds](https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml).

I would expect that, as having a continuously failing PR check condition removes automatically any value that condition could provide, unsafe PR tests should not be executed against saucelabs, so that the build color provides quick feedback on the quality of the code behind the PR (and we external contributors don't get crazy about how to fix the build :-) ).

TimvdLippe:
[...]

PR #3354, opened 2019-02-11
WIP - ci(): fix external PR builds.

alfonso-presa:
fixes: #3353

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,16 @@ matrix:
     - TEST_COMMAND="xvfb-run npm run test:integration"
 
   - node_js: '10'
+    if: type = pull_request
+    addons:
+      chrome: stable
+      firefox: latest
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -54,6 +54,16 @@ matrix:
     - TEST_COMMAND="xvfb-run npm run test:integration"
 
   - node_js: '10'
+    if: type = pull_request
+    addons:
+      chrome: stable
+      firefox: latest
+    env:
+    - SKIP_LOCAL_BROWSERS=true
```
TimvdLippe: If we are skipping local browsers here, then we don't need the addons here. That is the real bug: we expect the addons to be there, even though we explicitly skip them. As I said in the issue, we need to update `run-wct.sh` instead of fixing the Travis configuration.

Issue #3348, opened 2019-02-04
[bundler] Preserve absoluteness of paths


usergenic:
If a URL in a file contains an absolute path, we should honor the developer's intent and it should be emitted as an absolute path even if the target has moved due to bundling.

This is related to https://github.com/Polymer/tools/issues/1330

usergenic:
cc @esmael-e

usergenic:
Fixed in #3357

usergenic:
Published in 4.0.8

PR #3357, opened 2019-02-12
Absolute urls should remain absolute

usergenic:
Preserves the absoluteness of URLs in elements when bundling, where they were all being relativized and that's not fair.


usergenic:
Fixes #3348

usergenic:
Thx @aomarks

PR #3366, opened 2019-02-27
Upgraded to latest version of rollup

usergenic:
- Added option for bundler to let rollup treeshake.
 - Added a bundler test to demonstrate treeshake with PURE annotations.

### DIFF
--- a/packages/bundler/CHANGELOG.md
+++ b/packages/bundler/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
+* Upgraded Rollup to version 1.3.0.
+* Added `treeshake` option to enable Rollup to remove dead code when bundling.
 <!-- Add new, unreleased changes here. -->
 
 ## 4.0.8 - 2019-02-12
--- a/packages/bundler/src/es6-rewriter.ts
+++ b/packages/bundler/src/es6-rewriter.ts
@@ -152,24 +152,22 @@ export class Es6Rewriter {
       throw new Error(`Failed to bundle.  Rollup generated ${
           output.length} chunks or assets.  Expected 1.`);
     }
-    for (const chunkOrAsset of output) {
-      const rolledUpCode = chunkOrAsset.code || '';
-      // We have to force the extension of the URL to analyze here because
-      // inline es6 module document url is going to end in `.html` and the file
-      // would be incorrectly analyzed as an HTML document.
-      const rolledUpUrl = getFileExtension(url) === '.js' ?
-          url :
-          appendUrlPath(url, '_inline_es6_module.js');
-      const rolledUpDocument = await this.bundler.analyzeContents(
-          rolledUpUrl as ResolvedUrl, rolledUpCode);
-      const babelFile = assertIsJsDocument(rolledUpDocument).parsedDocument.ast;
-      this._rewriteExportStatements(url, babelFile);
-      this._rewriteImportStatements(url, babelFile);
-      this._deduplicateImportStatements(babelFile);
-      const {code: rewrittenCode} = serialize(babelFile);
-      return {code: rewrittenCode, map: undefined};
-    }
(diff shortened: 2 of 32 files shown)


usergenic:
cc @westbrook

packages/bundler/src/es6-rewriter.ts
```diff
@@ -139,28 +138,38 @@ export class Es6Rewriter {
...
-    this._rewriteImportStatements(url, babelFile);
-    this._deduplicateImportStatements(babelFile);
-    const {code: rewrittenCode} = serialize(babelFile);
-    return {code: rewrittenCode, map: undefined};
+
+    if (output.length !== 1) {
+      throw new Error(`Failed to bundle.  Rollup generated ${
+          output.length} chunks or assets.  Expected 1.`);
+    }
+    for (const chunkOrAsset of output) {
```
justinfagnani: do you need the loop if the length is 1?

packages/bundler/src/es6-rewriter.ts
```diff
@@ -139,28 +138,38 @@ export class Es6Rewriter {
...
+      const rolledUpDocument = await this.bundler.analyzeContents(
+          rolledUpUrl as ResolvedUrl, rolledUpCode);
+      const babelFile = assertIsJsDocument(rolledUpDocument).parsedDocument.ast;
+      this._rewriteExportStatements(url, babelFile);
+      this._rewriteImportStatements(url, babelFile);
+      this._deduplicateImportStatements(babelFile);
+      const {code: rewrittenCode} = serialize(babelFile);
+      return {code: rewrittenCode, map: undefined};
+    }
+    throw new Error(`Failed to bundle.  Rollup generated no chunks or assets.`);
```
justinfagnani: shouldn't this be redundant with the `output.length` check above?

packages/bundler/src/es6-rewriter.ts
usergenic: This is a holdover from when I wasn't sure if I needed to find or merge chunks, which it doesn't look like I will.  It does obscure the intent leaving it this way.  Removed the loop.  Thanks.

packages/bundler/src/es6-rewriter.ts
usergenic: Yeah but the for loop caused typescript to declare a possible undefined return.  This problem has gone away by removing the for loop and thus this throw is unneeded and now removed.

packages/bundler/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
+* Upgraded Rollup to version 1.3.0.
+* Added `treeshake` option to enable Rollup to remove dead code when bundling.
```
Westbrook: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

packages/cli/src/test/integration/integration_test.ts
```diff
@@ -54,8 +55,12 @@ async function serve(dirToServe: string, options: ServerOptions = {}) {
   const address = startResult.server.address();
   if (typeof address === 'string') {
     return `http://${address}`;
-  } else {
+  } else if (!isNull(address)) {
     return `http://${address.address}:${address.port}`;
+  } else {
+    // How you gonna serve without an address?  How is that even a server?  If a
+    // server can't respond to requests, is it *reall* a server?
```
Westbrook: Sarcasm is spelled with a `y` at the end here? I mean, "reallY".
```suggestion
    // server can't respond to requests, is it *really* a server?
```

usergenic:
@westbrook can you sign the CLA for your contribution of the Y in "really"?

Westbrook:
That's weird. I've signed them before... in this repo. Is there a link to do it again?

Also, sorry for so much trouble over one little letter.

usergenic:
@Westbrook I'm marking CLA: yes because that commit is immaterial and I don't know what googlebot is talking about.

Westbrook:
Love it!

PR #3367, opened 2019-02-28
Polymer treeshake

usergenic:
treeshake

Adds support for `treeshake` option in Polymer Project Config, Build and CLI now that it's in Bundler.

### DIFF
--- a/packages/build/src/bundle.ts
+++ b/packages/build/src/bundle.ts
@@ -45,41 +45,22 @@ export class BuildBundler extends AsyncTransformStream<File, File> {
 
     this._buildAnalyzer = buildAnalyzer;
 
-    const {
-      analyzer,
-      excludes,
-      inlineCss,
-      inlineScripts,
-      rewriteUrlsInTemplates,
-      sourcemaps,
-      stripComments,
-      urlMapper
-    } = options;
-    let {strategy} = options;
+    const bundlerOptions = {...options};
 
-    const urlLoader =
-        new FileMapUrlLoader(this.files, analyzer || buildAnalyzer.analyzer);
-    const forkedAnalyzer =
-        (analyzer || buildAnalyzer.analyzer)._fork({urlLoader});
+    const urlLoader = new FileMapUrlLoader(
+        this.files, bundlerOptions.analyzer || buildAnalyzer.analyzer);
+    bundlerOptions.analyzer =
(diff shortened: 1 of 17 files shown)


usergenic:
cc @westbrook

usergenic:
*Note that `treeshake` is still false by default so this should have no effect on existing project builds.*

packages/build/src/bundle.ts
```diff
@@ -45,41 +45,22 @@ export class BuildBundler extends AsyncTransformStream<File, File> {
...
-
-    const urlLoader =
-        new FileMapUrlLoader(this.files, analyzer || buildAnalyzer.analyzer);
-    const forkedAnalyzer =
-        (analyzer || buildAnalyzer.analyzer)._fork({urlLoader});
-
-    if (strategy === undefined && this.config.shell !== undefined) {
-      strategy =
-          generateShellMergeStrategy(forkedAnalyzer.resolveUrl(urlFromPath(
+    const bundlerOptions = {...options};
```
Westbrook: I'll try not to commit any code this time, but clarification question, what's the benefit of having a copy of the `options` object here rather than just working with the one supplied?

usergenic: I don't want the function to mutate the object handed to it.  I swap the analyzer for a forked version.

packages/cli/test/fixtures/build-modules/expected/bundled-amd/shared_bundle_1.js
```diff
@@ -19,8 +19,12 @@ define(["exports"], function (_exports) {
   _exports.$mod2 = mod2$1;
   const mod1 = 'mod1';
   _exports.mod1 = mod1;
+  sideEffectFunction();
+  /*@__PURE__*/
```
aomarks: What is @__PURE__ ?

usergenic: This is an annotation that was introduced initially by UglifyJS to mark "pure" functions (having no side effects) so they can be safely removed unless their result is actually used.

usergenic: Rollup honors it and does not remove function calls unless it is certain they have no side effect.

Taewa:
Thanks guys!! :)
Available on polymer2 as well?

PR #3379, opened 2019-03-07
Support option allowOrigin

quentinmazoua:
Support option allowOrigin


usergenic:
Working backwards I merged an identical PR #3439 from @T99Rots which did same thing but thanks for posting this fix first, @quentinmazoua

PR #3382, opened 2019-03-13
Fix Firefox support on iOS

plinss:
Currently prpl-server sends the wrong build to Firefox on iOS. This patch changes the browser detection so that all browsers on iOS are treated as Mobile Safari, not just Chrome.


plinss:
I signed it!

plinss:
This is still an issue.

A one-line fix PR open for over a year makes one doubt the commitment to this project... or is it just an unwillingness to support browsers other than Chrome?

plinss:
Two years?

PR #3383, opened 2019-03-21
docs: update bundle-manifest's url in Readme

armujahid:
I have updated bundle-manifest's url because it is currently pointing to outdated mirror https://github.com/Polymer/polymer-bundler/.


armujahid:
@usergenic Kindly review this. Its a simple URL change in docs.

PR #3391, opened 2019-04-01
Upgrade to selenium standalone server 3.141.59

nelsonsilva:
This should fix https://github.com/Polymer/tools/issues/405


usergenic:
Trying this out with latest JDKs. Thanks, @nelsonsilva

usergenic:
Will be releasing this ASAP.  Thanks again.

Issue #3403, opened 2019-04-19
[polymer-serve] Dynamic imports fail in Firefox / AMD transformed browsers


btopro:
This is related to #3402 most likely. When using `polymer serve` to view an asset that dynamically imports via bare import syntax, it fails to resolve correctly.

```js
import("@polymer/paper-button/paper-button.js");
```

This is most likely due to a timing issue in the babel transform related to doing this as it was the case in the similar issue with build.

e111077:
Firefox 67 shipped yesterday (May 21, 2019) which provides native support for dynamic imports. Updating `browser-capabilities` to sidestep this issue.

btopro:
I'm surprised the bare import resolution didn't solve this but good call, then it'll at least work w/ the latest and greatest polyserve locally (which is probably fine)

Issue #3402, opened 2019-04-18
transformModulesToAmd + unbundled option fails to resolve bare imports


btopro:
the transformModulesToAmd option in build routinse in Polymer build does not correctly resolve bare dynamic imports. This is visible in firefox where it will convert it to a require statement but not actually alter the path, causing it to try and be relative to where ever the current file is.
```js
TypeError: "Failed to fetch https://cdn.waxam.io/build/es5-amd/node_modules/@lrnwebcomponents/haxcms-elements/lib/core/@lrnwebcomponents/haxcms-elements/lib/core/haxcms-editor-builder.js"
```
in this example: `https://cdn.waxam.io/build/es5-amd/node_modules/@lrnwebcomponents/haxcms-elements/lib/core/haxcms-site-builder.js` will call `import("@lrnwebcomponents/haxcms-elements/lib/core/haxcms-editor-builder.js")`. This works fine in ES6 but when it goes to path resolve in ES5 you get what you see above as opposed to any kind of `../../../../` being appended to make the path relative accurately

btopro:
here's my polymer.json https://github.com/elmsln/HAXcms/blob/master/polymer.json#L47 which is leveraged to kick things over to that CDN.

btopro:
TLDR;
- Good: transformModulesToAmd + bundled + dynamic imports
- Good: unbundled + dynamic imports
- Broken: transformModulesToAmd + unbundled + dynamic imports

This is an issue with unbundled AMD module compiling.
steps to reproduce:
- Test repo: https://github.com/btopro/lit-test
- run `polymer build`

This repo's output will WORK for the one marked unbroken and will break for the one marked broken. The only difference is the bundling. What's breaking appears to be a combination of transformModulesToAmd when using the unbundled flag. Dynamic imports Unbundled works fine without the AMD transform. AMD works fine without dynamic imports.

btopro:
thanks to @Westbrook for making the repo to confirm I wasn't crazy

usergenic:
Just published polymer-cli@1.9.9 and polymer-build@3.1.4 which incorporate your PR.

Please install and verify fix for your use case.  Reopen this issue if you have continued trouble.  :)

btopro:
installed to double check and yes this appears to have fixed it. Thank you for accepting so quickly!

PR #3404, opened 2019-04-19
[polymer-build] #3402 resolveBareSpecifiers must be a Program

btopro:
Babel's life cycle appears to make the following combination not possible with the current configuration of polymer's build process.
- `"moduleResolution": "node"`
- `"transformModulesToAmd": true`
- `"bundle": false,`

## What ends up happening now:
- Babel processes the tree and transforms `ast` for AMD compilation and/or ES5 syntax in the Program level callback
- This happens PRIOR to processing `babel-plugin-bare-specifiers` `resolveBareSpecifiers` export because it is running in a `CallExpression` which fires later
- The `CallExpression` seeks out things marked 'Import' but because the compiling has already happened, it can't find everything
- Dynamic imports get borked as a result, not being correctly resolved to their actual location in node_modules because it doesn't process it as an Object tree but instead as compiled code.

## This change fixes this by..
This shifts it to process at the Program level and traverse the path to execute `CallExpression`'s. This should ensure that the timing is correct so that the Objects can be resolved to their correct node_modules / relative location, prior to the ES5 / AMD transformations processing the tree into text.

## Gains via this PR
- Users will be able to user polymer.json as expected
- Polymer build will correctly resolve dynamic imports regardless of compiling or module resolution used

## Outstanding issues
#3403 is most likely related to this but I haven't been able to see what the issue is as to why this wouldn't overlap.

### DIFF
--- a/packages/build/src/babel-plugin-bare-specifiers.ts
+++ b/packages/build/src/babel-plugin-bare-specifiers.ts
@@ -14,7 +14,7 @@
 
 import dynamicImportSyntax from '@babel/plugin-syntax-dynamic-import';
 import {NodePath} from '@babel/traverse';
-import {CallExpression, ExportAllDeclaration, ExportNamedDeclaration, ImportDeclaration} from 'babel-types';
+import {CallExpression, ExportAllDeclaration, ExportNamedDeclaration, ImportDeclaration, Program} from 'babel-types';
 import {resolve} from 'polymer-analyzer/lib/javascript/resolve-specifier-node';
 
 const isPathSpecifier = (s: string) => /^\.{0,2}\//.test(s);
--- a/packages/build/src/babel-plugin-bare-specifiers.ts
+++ b/packages/build/src/babel-plugin-bare-specifiers.ts
@@ -35,23 +35,26 @@ export const resolveBareSpecifiers = (
   inherits: dynamicImportSyntax,
 
   visitor: {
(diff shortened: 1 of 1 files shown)


LarsDenBakker:
Running into this issue as well. Thanks for making the PR!

Westbrook:
@usergenic Using this update locally solved this issue for me as well. If you like what you see, too, I hope we can get a merge and release scheduled for early next week.

usergenic:
Hey guys - I'm just catching up after holiday weekend.  Saw your slack thread and then came over here.  Am reviewing now.

I'm actually not as familiar with the build babel transforms and my first reaction is, why do we do compilation *before* resolving these specifiers; seems like we should be resolving specifiers first...

Anyways looking at this now.  Thanks for this work, @btopro!

usergenic:
I'm still working on writing a test to repro the problem and include it as a regression test, but if you want to check to see that this solves issue for you, I've got a simple PR that reordered the specifier transform plugin https://github.com/Polymer/tools/pull/3405

usergenic:
Our dynamic import AMD transform is causing the problem because it is written as a Program visitor that does its own traversal, essentially making it always perform its own CallExpression transforms first.  After a couple hours of attempting to refactor etc, it turns out your approach here is the most reasonable solution, so I'm going to go with this one.  Great job :)

PR #3405, opened 2019-04-23
Resolve bare specifiers PRIOR to compilation.

usergenic:
There's an issue about specifiers NOT being transformed when the AMD transform is involved as reported in https://github.com/Polymer/tools/issues/3402 https://github.com/Polymer/tools/issues/3403

This PR is a possible alternative to https://github.com/Polymer/tools/pull/3404, inferring from that work that the ordering of babel transform plugins was at issue.  This simply moves specifier resolution to the beginning of the js-transform sequence.

### DIFF
--- a/packages/build/src/babel-plugin-dynamic-import-amd.ts
+++ b/packages/build/src/babel-plugin-dynamic-import-amd.ts
@@ -71,9 +71,9 @@ export const dynamicImportAmd = {
           // assume is running next will rewrite `require` from a function to a
           // module object with the function at `default`.
           importPath.replaceWith(ast`(
-          new Promise((res, rej) => ${requireId}.default([${
+            new Promise((res, rej) => ${requireId}.default([${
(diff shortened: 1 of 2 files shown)


usergenic:
cc @btopro @westbrook

btopro:
were you able to confirm this fixes the transform issue? I believe I tried this first initially and the issue is that it doesn't process the plugins in the same order. It seems like `CallExpression` are processed after `Program`, so the other PR changes the transform statement to be a looping `Program` as opposed to single `CallExpression` that runs post-transformation.

usergenic:
@btopro OH the nature of the problem comes from the fact that our babel-plugin-dynamic-import-amd.ts is doing a path.traverse in its own Program visitor 🤦‍♂️

I'm going to try and fix that right now too.

usergenic:
@btopro I'm going to go with your solution in #3404

PR #3436, opened 2019-06-05
Upgrade selenium standalone server to support Java 10+

usergenic:
Tested this change locally using openjdk@1.13.0 which did not work until this change was introduced and now it works.  By "did not work" I mean that selenium threw a bunch of exceptions and wasn't useable.


usergenic:
https://github.com/Polymer/tools/pull/3391 was the source of this so Googlebot CLA check is being dumb.

nelsonsilva:
I'm ok with their commits being contributed to this project, feel free to update the cla label

PR #3438, opened 2019-06-06
Prepare releases of wct-local and polymer-cli

usergenic:
- [ ] wct-local@1.0.1
- [ ] polymer-cli@1.9.10


usergenic:
ty! @aomarks

PR #3439, opened 2019-06-13
fixed --allow-origin switch not being added to serverOptions

T99Rots:
Currently when using polyserve with the polymer-cli the --alow-origin switch does not work, this is because --allow-origin is not being added to the serverOptions, i now added it in and tested it and now it's working as it's supposed to.


T99Rots:
I signed it!

usergenic:
Thanks for this PR!  Merging into a branch to add CHANGELOG.

PR #3443, opened 2019-07-03
Added --allow-origin to the polymer serve command

usergenic:
--allow-origin was supported by polyserve, but we never brought that flag into the CLI.  This PR brings that in.  Thanks to @T99rots


usergenic:
Hey @googlebot you approved his CLA in https://github.com/Polymer/tools/pull/3439 so I'm going to ignore you:

Screen Shot 2019-07-03 at 2 38 29 PM

usergenic:
Thanks also to @quentinmazoua for an earlier PR #3379 which added this first.

PR #623, opened 2018-08-06
cli: update chalk dependency

43081j:
I was in there for #612 so figured i'd split out the chalk update i did as part of my debugging.

We don't really need this but it does mean we're depending on the same version as `command-line-usage` so there's no more mismatch funkiness going on.

PR #3444, opened 2019-07-03
Updated chalk to latest version; implemented the updates from #623

usergenic:
Includes newer version of chalk and updated package-locks all-around.

Resolving conflicts in package-lock.json from #623 was not convenient without access to branch so this was most expedient way to update the library.

Thanks to @43081j

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1021,12 +1021,12 @@
       "dev": true
     },
     "@octokit/endpoint": {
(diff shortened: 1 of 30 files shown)


usergenic:
Just to confirm chalk updates don't blow up the polymer logo as it did after the last upgrade:
Screen Shot 2019-07-03 at 3 28 48 PM

PR #3445, opened 2019-07-03
[wct-local] Added javaArgs to plugin options.

usergenic:
Allows passing of arguments to java invocation.


usergenic:
CLA was already signed for prior PR https://github.com/Polymer/tools/pull/3309

PR #3454, opened 2019-08-26
Open the url in browser when user serves their polymer app using the open library

aviskarkc10:
1. Install [sindresorhus/open](https://github.com/sindresorhus/open/).
2. Open the browser after user runs `polymer serve`


aviskarkc10:
@googlebot I signed it

aviskarkc10:
Oops! My bad. Just found out there is an `--open` flag. Sorry :smile:

PR #3460, opened 2019-10-01
docs: add lerna badge

armujahid:
Polymer tools are organized as a monorepo using [Lerna](https://lernajs.io). So I think we should either add lerna badge
[lerna](https://lerna.js.org/) or some kind of acknowledgment in Readme (If addition of badge is not acceptable).


armujahid:
This issue shouldn't be closed because related PR is already open.

PR #3479, opened 2019-12-18
Fix: Polymer CLI tests not running after "npm install"

rgladwell:


rgladwell:
@googlebot I signed it!

PR #3496, opened 2020-04-12
Fix ineffective --component-url serve cli option

### LINKED ISSUE #3495: Polymer serve's --component-url option has no effect
Polymer serve's `--component-url` option has no effect on the actual component url used, which is always set to `components`.

**To reproduce**
Run
```bash
$ polymer serve --component-url <some-component-url>
```
The output is always:
```
> info:   Files in this directory are available under the following URLs
       applications: http://localhost:8081
       reusable components: http://localhost:8081/components/<root-name>/
                                                  ^^^^^^^^^^
```

**Expected behavior**
The reusable components should be served under `http://localhost:8081/<some-component-url>/<root-name>/`.

---

This makes polymer cli's server practically unusable in any project where the components are (understandably) placed in a folder named `components`.

umbopepato:
Fixes #3495: add `componentUrl` to the server options passed to
polyserve, pass `serverOptions` instead of `options` to
`getServerUrls()`

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -69,6 +69,7 @@ export class ServeCommand implements Command {
       npm: config.npm,
       moduleResolution: config.moduleResolution,
       componentDir: config.componentDir,
+      componentUrl: options['component-url'],
       packageName: options['package-name'],
       protocol: options['protocol'],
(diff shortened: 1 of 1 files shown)


umbopepato:
@googlebot I signed it!

PR #3537, opened 2020-10-03
dom5: update to latest parse5

43081j:
This is a fairly big change because parse5's types are pretty terrible. im trying to fix that in definitelytyped but it'll take time, so meanwhile i import all the default interfaces here as what they should be (e.g. `DefaultTreeNode` as `Node`).

cc @aomarks

### DIFF
--- a/packages/dom5/package-lock.json
+++ b/packages/dom5/package-lock.json
@@ -1,53 +1,111 @@
 {
-	"name": "dom5",
-	"version": "3.0.1",
-	"lockfileVersion": 1,
-	"requires": true,
-	"dependencies": {
-		"@types/clone": {
-			"version": "0.1.30",
-			"resolved": "https://registry.npmjs.org/@types/clone/-/clone-0.1.30.tgz",
-			"integrity": "sha1-5zZWSMG0ITalnH1QQGN7O1yDthQ=",
-			"dev": true
-		},
-		"@types/parse5": {
-			"version": "5.0.3",
-			"resolved": "https://registry.npmjs.org/@types/parse5/-/parse5-5.0.3.tgz",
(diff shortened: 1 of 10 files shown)


43081j:
this may depend on me figuring out DefinitelyTyped/DefinitelyTyped#48402 as these types are just unbearable. 

we have to cast almost every node here so whats the point in the types? its so bad... even a childNode of a node isn't considered a `ChildNode`... 

im afraid if i just fix all this up we'll be shipping a dom5 release that depends on these completely insane types in such a way that people can't easily consume it

43081j:
i've just pushed a commit here that will fail our build...

we are dependent on my parse5 PR over at the definitelytyped repo at the min. so i've gone ahead and fixed up this branch assuming those types are available.

once my PR lands (if ever), ill update this branch of dom5 too to point at it.

rictic (commented):
These changes look good to me once the typings are ready and CI is happy.

43081j:
Awesome, thanks for taking a look at it. It's a bit of a blast from the past I know. 

Hopefully I can get the DT pull request sorted soon

43081j:
@rictic i finally got the types merged..

however, it doesn't look like i can get this working without also updating polymer-analyzer to use newer parse5. it seems the analyzer fails to build because it pulls in a different parse5 than it wants. im not exactly sure why that happens, monorepo witchcraft

maybe you could also bounce the idea around the team of dom5 being moved into its own repo at polymerlabs or lit-html? as @usergenic [commented here](https://github.com/Polymer/lit-html/blob/lit-next/packages/lit-ssr/src/lib/util/parse5-utils.ts) there's also a copy in lit to work around the lack of this exact PR.

PR #3611, opened 2022-01-04
Pin to a version of launchpad without CVE-2021-23330

rictic:


AndrewJakubowicz (commented):
LGTM! Thank you! Is CI failure related?

AndrewJakubowicz (approved):

web-padawan:
@rictic FYI, this PR is likely to make it impossible to install `web-component-tester` with `yarn` due to error:

```
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads git://github.com/418sec/launchpad.git
Directory: /Users/serhii/vaadin/vaadin-dialog
Output:
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

As the new commits are not yet release, I suppose there is some other third party dependency causing it for me 🤔

rictic:
the new commits are released as part of wct-local v2.1.6

as a temporary workaround you can pin to v2.1.5

it sounds like the https version of the url should work, let me try that

web-padawan:
@rictic I tried to pin using npm 8.3.0 overrides feature but WCT still fails with a launchpad error:

```
Run xvfb-run -s '-screen 0 1024x768x24' npm test

> @vaadin/vaadin-item@2.3.0 test
> wct

(node:2206) [DEP0131] DeprecationWarning: The legacy HTTP parser is deprecated.
step: loadPlugins
step: configure
hook: configure

Invalid characters inside the path to the browser

Invalid characters inside the path to the browser

TypeError: Cannot read property 'match' of undefined

Error: Process completed with exit code 1.
```

Example: https://github.com/vaadin/vaadin-item/runs/4778604110?check_suite_focus=true

rictic:
Aiming to release a fix by end of day

web-padawan:
Thanks. Have you encountered the `Invalid characters inside the path to the browser` error BTW?

web-padawan:
OK, so basically here is what we get in GitHub Actions for running local browsers:

```
[{"name":"chrome","path":"/usr/bin/google-chrome","binPath":"/usr/bin/google-chrome"},{"name":"chromium","path":"/usr/bin/chromium-browser","binPath":"/usr/bin/chromium-browser"},{"name":"firefox","version":"Mozilla Firefox 95.0","path":"/usr/bin/firefox","binPath":"/usr/bin/firefox"},{"name":"phantom","version":"2.1.1","path":"/usr/local/bin/phantomjs","binPath":"/usr/local/bin/phantomjs"}]
```

This is tested against the [RegExp](https://github.com/418sec/launchpad/blob/de5aca11dc16a8e530195281c77614bdbb08e7be/lib/local/version.js#L12) added to launchpad `var filter = /[`!@#$%^&*()_+\-=\[\]{};'"|,<>?~]/;`

And then it returns `true` because of dash character present in `"/usr/bin/google-chrome"` 🤦‍♂️

rictic:
That one is new to me

rictic:
oh no

rictic:
If you want to fork the fork of launchpad and apply a fix for the fix I'd be happy to pin to that

web-padawan:
Looks like I have to do that - will ping you here when it's done 😰   
Note, I haven't checked the CVE but I suppose `-` alone is very unlikely to cause security issues.

rictic:
Would it be possible to be more strict, e.g. by only allowing a hyphen before the first space?

web-padawan:
Here is a commit with hyphen removed https://github.com/web-padawan/launchpad/commit/86415a9fedd5cdfd0426806e4a032a63c845105c

> Would it be possible to be more strict, e.g. by only allowing a hyphen before the first space?

Not if I can quickly come up with a correct RegExp for it, but I will try

web-padawan:
Maybe we can explicitly check for `"google-chrome"` and `"chromium-browser"` and only allow dash there 🤔

rictic:
Any other browsers that are written like that? microsoft-edge? firefox is just firefox and safari is just safari.

Works for me

web-padawan:
OK something like this should work I guess
https://github.com/web-padawan/launchpad/commit/fdd994d089572e2c2152d70cc74abf497d08d5b7

web-padawan:
@rictic The above commit works https://github.com/vaadin/vaadin-item/runs/4779649002?check_suite_focus=true

rictic:
@web-padawan published as wct-local 2.1.7!

web-padawan:
Thanks a lot for your quick feedback! Confirmed that this version works in our web components CI builds 🚀

PR #3624, opened 2022-01-11
Use https instead of git URL

rictic:
HTTPS should always be encrypted, while git might not be, and github no longer supports unencrypted git connections. See https://github.blog/2021-09-01-improving-git-protocol-security-github/


rictic:
See https://github.com/Polymer/tools/pull/3611#issuecomment-1009735367

web-padawan (commented):
Tested with yarn and it solves the problem 👍

AndrewJakubowicz (approved):
LGTM! Thanks for the manual test web-padawan!

